PR #1, opened 2018-02-12
Make the geocoder return more options for address messages to use,

fosJoddie:
When reverse geocoding coordinates, return not only the formated address, but also the individual components to allow for more complex message DTS like `{{city}}`

Fix error in address lookup for raid/monster

Added DTS template keywords:

streetNumber - the street number
streetName - name of the street
zipcode - zip code if applicable
country - country 
countryCode - two letter country code
city - the city
state - the US state name (possibly others, don't know)
stateCode: two letter state code (unknown what this does for non us addresses, probably blank)

tested:
yes?

### DIFF
--- a/app/src/geo/google.js
+++ b/app/src/geo/google.js
@@ -3,12 +3,11 @@ let config = require('config');
 let inside = require('point-in-polygon');
 let _ = require('lodash');
 let NodeGeocoder = require('node-geocoder');
+
 let geocoder = NodeGeocoder({
     provider: 'google',
     httpAdapter: 'https',
-    apiKey: config.gmaps.key,
-    formatter: 'string',
-    formatterPattern: config.locale.addressformat
+    apiKey: config.gmaps.key
 });
 let getCoords = NodeGeocoder({
     provider: 'google',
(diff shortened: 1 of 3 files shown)


app/src/geo/google.js
```diff
@@ -27,10 +26,43 @@ module.exports = {
     },
 
     getAddress: function(location, callback) {
-        geocoder.reverse(location, function(err, res) {
+        geocoder.reverse(location, function(err, data) {
```
KartulUdus: why rename res to data here 🤔

app/src/geo/google.js
```diff
@@ -3,12 +3,11 @@ let config = require('config');
 let inside = require('point-in-polygon');
 let _ = require('lodash');
 let NodeGeocoder = require('node-geocoder');
+
 let geocoder = NodeGeocoder({
```
KartulUdus: This makes geocoder === getCoords

Could you make the geolocate && getAddress both use geocoder too, to not have duplicate declarations

app/src/geo/google.js
fosJoddie: data is the raw data from geocoder, while res is what is returned to the callback of getAddress. 

It was an attempt to clear up the dataflow in my own head.

PR #4, opened 2018-02-20
Store geocodes in a time limited cache to avoid repeated lookups

fosJoddie:
Add the use of a persistent cache to save address lookups for 30 days (max allowed by google) so that we do not have to worry about key expiration because of lookups to the same addresses.

adds a new requirement, so a `npm install` is required after upgrade.

Also not sure on the "global" method in google.js - cause I don't know node.js enough to understand where such utility functions should live.

### DIFF
--- a/app/src/geo/google.js
+++ b/app/src/geo/google.js
@@ -3,6 +3,7 @@ let config = require('config');
 let inside = require('point-in-polygon');
 let _ = require('lodash');
 let NodeGeocoder = require('node-geocoder');
+let pcache = require('persistent-cache');
 
 let geocoder = NodeGeocoder({
     provider: 'google',
@@ -12,6 +13,12 @@ let geocoder = NodeGeocoder({
 
 let opengeocoder = NodeGeocoder({
 	provider: 'openstreetmap'
+});
+
+let addrCache = pcache({
+    base: '.cache',
+    name: 'addrCache',
+    duration: 30 * 24 * 3600 * 1000 //30 days is what google allows
 });
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


Pogodenhelder:
trying out this PR right now.

app/src/geo/google.js
```diff
@@ -21,43 +65,27 @@ module.exports = {
     },
 
     getAddress: function(location, callback) {
-        geocoder.reverse(location, function(err, geocodeResult) {
-            if (err) log.error(err);
+        var cacheKey = location.lat + "-" + location.lon;
```
KartulUdus: don't use string + concat like this please 

use 
let cacheKey = `${location.lat}-${location.lon}`;

instead

app/src/geo/google.js
```diff
@@ -21,43 +65,27 @@ module.exports = {
...
+        addrCache.get(cacheKey, function(err, addr) {
+            if(err) {
+                log.error(err);
```
KartulUdus: not need for parantheses 

`if(err) log.error(err);` 
will do

app/src/geo/google.js
```diff
@@ -21,43 +65,27 @@ module.exports = {
...
-				res.stateCode = '';
-			}
+            if(addr) {
+                parseAddr(err, addr, callback);
+                return;
+            }
 
-            callback(err, res);
+            geocoder.reverse(location, function(gerr, geocodeResult) {
+                if (gerr) log.error(gerr);
```
KartulUdus: why gerr 🤔

fosJoddie: It feels dirty outscoping err from the outer promises. Is that a bad thing?

fosJoddie:
did the needful, I hope

PR #5, opened 2018-02-20
Support for embed color by perfection for IV alerts

fosJoddie:
Add support to use iv color instead of type color for monster messages

new dts `{{ivcolor}}` will color alarms depending on how good a pokemons IV is:

gray: less than 25 %
white: 25 to 50% 
green: 50 - 82%
blue: 82 to 90 %
purple: 90 to 100%
orange: 100%


Alderon86:
💯 could be red to match emoji style

fosJoddie:
they follow the world of warcraft loot color, because they are imprinted in many games.

And they follow the rule of "don't let red be best cause Valor would have a field day"

Alderon86:
Valor 4 Life! :D

app/src/filter.js
```diff
@@ -354,3 +356,27 @@ function sendDMAlarm(message, human, e, map) {
         })
     } else log.warn(`Tried to send message to ID ${human}, but didn't find it`);
 }
+
+
+function findRarityColor(iv) {
```
KartulUdus: perhaps rarity is misleading, a pidgey can be 💯 but is not rare

{{ivcolor}} perhaps

app/src/filter.js
```diff
@@ -354,3 +356,27 @@ function sendDMAlarm(message, human, e, map) {
...
+    if(!iv || iv < 25) {
```
KartulUdus: `!iv` won't happen because missing iv === -1 (%)

config/dts.json.example
```diff
@@ -4,7 +4,7 @@
       "title": "a wild {{{name}}} has appeared!",
       "description": "It weighs {{weight}}kg and will despawn at {{time}}, you have {{tthm}}m {{tths}}s left \nPerfection: **{{iv}}%** , ({{atk}}/{{def}}/{{sta}}) \n{{move1}}|{{move2}} Level:{{level}} (cp:**{{cp}}**)\naddress: {{addr}} ",
       "url": "{{{rocketmap}}}",
-      "color": "{{color}}",
+      "color": "{{rarityColor}}",
```
KartulUdus: I like that you can get raritycolor for IV and typecolor for noIV 👍

fosJoddie:
Changed to `ivcolor`

moved colors into the default.json config so people can change color themselves.

Added docs. Heck knows if I made a good table or not

PR #11, opened 2018-02-23
Embed color by iv quality

fosJoddie:
add `{{ivcolor}}` as moustache DTS for monsters with IV information. Colors the embeds by iv quality instead of `color` that is color by type.

Config moved to it's own json with fallback if people upgrade and don't add it in their config files


docs/config.md
```diff
@@ -38,7 +38,15 @@ This config file consist of different objects for different configurations.
     "host": "127.0.0.1",
     "port": "3031",
     "imgurl": "https://raw.githubusercontent.com/KartulUdus/PoracleJS/master/app/src/util/images/",
-    "max_pokemon": 326
+    "max_pokemon": 326,
```
KartulUdus: Could this be in the config.discord object please

config/default.json.example
```diff
@@ -21,7 +29,7 @@
     "addressformat": "%S %n",
     "commandMonstersJson": "../util/monsters",
     "monstersJson": "./util/monsters",
-    "movesJson": "./util/moves"
+    "movesJson": "./util/moves",
```
KartulUdus: 🤔 last value in object added comma, why dat?

fosJoddie:
Changes as requested

PR #23, opened 2019-02-13
Add config option to only send alert about monsters with more than X sec to hidden

fosJoddie:
## Description
To prevent alerts with say 1 minute remaining before a monster goes away, make it configurable to ignore anything with less than X seconds until hidden

## Motivation and Context
Solves the annoying problem of a good mon showing up with 1 min left

## How Has This Been Tested?
I showed it to a band of traveling merchants and they aproved.

## Types of changes

## Checklist:


KartulUdus:
Very nice!

One note is to bump the version number (middle digit as it's a new feature)
  "version": "2.8.10", with current ver 😄

PR #39, opened 2019-05-21
Parameterize insert SQL

fosJoddie:
100% untested - just to show my thoughts

## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/controller.js
+++ b/src/controllers/controller.js
@@ -255,13 +255,15 @@ class Controller {
 	async insertOrUpdateQuery(table, columns, values) {
 
 		const cols = columns.join(', ')
-		const multiValues = values.map(x => x.map(y => (typeof y === 'boolean' ? y : `'${y}'`)).join()).join('), \n(')
+		const multiValues = values.map(x => x.map(y => '?').join()).join('), \n(')
+		let params  = [];
+		values.forEach( (x) => { x.forEach( (y) => { params.push(y)})});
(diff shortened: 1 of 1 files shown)


src/controllers/controller.js
```diff
@@ -255,13 +255,15 @@ class Controller {
 	async insertOrUpdateQuery(table, columns, values) {
 
 		const cols = columns.join(', ')
-		const multiValues = values.map(x => x.map(y => (typeof y === 'boolean' ? y : `'${y}'`)).join()).join('), \n(')
+		const multiValues = values.map(x => x.map(y => '?').join()).join('), \n(')
```
KartulUdus: I believe the second .map() is useless as always returns '?'

src/controllers/controller.js
```diff
@@ -255,13 +255,15 @@ class Controller {
...
+		values.forEach( (x) => { x.forEach( (y) => { params.push(y)})});
 		const duplicate = columns.map(x => `\`${x}\`=VALUES(\`${x}\`)`).join(', ')
 		const query = `INSERT INTO ${table} (${cols})
-                      VALUES (${multiValues})
-                      ON DUPLICATE KEY UPDATE ${duplicate}`
+			      VALUES (${multiValues})
+			      ON DUPLICATE KEY UPDATE ${duplicate}`
 		return new Promise((resolve, reject) => {
-			this.db.query(query)
+		    this.db.query(query, params)
```
KartulUdus: params only has 1 set of valuses; however, on duplicate key update it would still go by string replacement

PR #40, opened 2019-05-29
Updating raid.js to fix forms for raids

t00fan:
Updated 2 lines to convert form id to form name.


KartulUdus:
Totally missed when this was created :D 

👍

PR #43, opened 2019-07-24
WIP Support pokestop webhook messages and incident alerts

fosJoddie:
### WARNING
This PR will require you to change your DTS with new stuff, check out "incident" section in the updated dts examples.

## Description
Add support for receiving pokestop webhooks and handle the incidents that come there. 
Team Rocket Incidents can be tracked with `!incident` command in a channel or PM, and will either react to incidents in your area(s) or based on distance from location if distance has been set like`!incident d5000`.

DTS names are `name` for pokestop name and `imgurl` for pokestop url. you have your normal tth and time for the EXPIRATION/END time and thats pretty much it. We do not know reward or anything like that.

For now there is ONLY support for MAD messages, since i do not have a RDM setup. If someone send me the payload for RDM I'll retrofit that in. Probably.

There's no way to see what incidents you track with `!tracked` - it has to be added in later.

## Motivation and Context
Because I wanna beat Team Rocket

## How Has This Been Tested?
Limited local testing on white bunnies and yellow chickens.

## Types of changes

## Checklist:

### DIFF
--- a/config/examples/dts.json.example
+++ b/config/examples/dts.json.example
@@ -354,7 +354,7 @@
         }
       }
     }
-  }
+  },
   "greeting": {
     "embed": {
       "title": "Welcome",
--- a/config/examples/dts.json.example
+++ b/config/examples/dts.json.example
@@ -381,8 +381,8 @@
           "value": "`{{prefix}}quest porygon pikachu poke ball d500 `: Any arguments are optional, this command would alert you about Quests obtainable within 500m of your location with porygon, pikachu or pokeballs as rewards \n `{{prefix}}remove quest all items` Removes tracking for all item based quests. Can also use `all pokemon` or `stardust`"
         },
         {
-          "name": "Incident tracking commands",
-          "value": "`{{prefix}}incident template3 d500 `: Any arguments are optional, this command would alert you about Team Rocket Incidents within 500m of your location\n `{{prefix}}incident remove` Removes tracking for all Team Rocket Incidents."
+          "name": "Invasion tracking commands",
+          "value": "`{{prefix}}invasion template3 d500 dragon mixed`: Any arguments are optional, this command would alert you about Team Rocket Incidents within 500m of your location if the grunt type was mixed or dragon. You can use any pokemon type name.\n `{{prefix}}invasion remove` Removes tracking for all Team Rocket Incidents."
         }
       ]
     }
--- a/config/examples/dts.json.example
+++ b/config/examples/dts.json.example
@@ -302,6 +302,59 @@
       }
     }
   },
+  "incident": {
+    "1": {
+      "embed": {
+        "author": {
+          "name": "R {{flagemoji}} {{lat}}, {{lon}} until: {{time}}, {{tthm}}m {{tths}}s",
+          "url": "{{{mapurl}}}",
+          "icon_url": "{{{imgurl}}}"
+        }
+      }
+    },
+    "2": {
+      "embed": {
+        "title": "{{name}}: incident, \n {{countryCode}} {{flagemoji}} {{addr}} {{lat}}, {{lon}} )",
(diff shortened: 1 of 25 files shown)


config/examples/dts.json.example
```diff
@@ -302,6 +302,59 @@
...
+        "description": "[Google Maps]({{{mapurl}}}) [Apple Maps]({{{applemap}}})",
+        "author": {
+          "name": "end: {{time}} in {{tthm}}m {{tths}}s",
+          "icon_url": "{{{imgurl}}}"
+        }
+      }
+    },
+    "3": {
+      "embed": {
+        "title": "Team Rocket incident has started! at {{ame}}",
```
Kneckter: Typo "ame" should be "name". Just a small fix for anyone testing.

src/controllers/incident.js
```diff
@@ -0,0 +1,180 @@
...
+				}
+				case 'mapbox': {
+					data.staticmap = `https://api.mapbox.com/styles/v1/mapbox/streets-v10/static/url-https%3A%2F%2Fi.imgur.com%2FMK4NUzI.png(${data.longitude},${data.latitude})/${data.longitude},${data.latitude},${config.geocoding.zoom},0,0/${config.geocoding.width}x${config.geocoding.height}?access_token=${_.sample(config.geocoding.staticKey)}`
+					break
+				}
+				default: {
+					data.staticmap = ''
+				}
+			}
+            if(data.name) data.gym_name = data.name;
```
Kneckter: I don't think this is used since we are not using gym_name.

src/controllers/incident.js
```diff
@@ -0,0 +1,180 @@
...
+			let incidentExpiration = data.incident_expiration ? data.incident_expiration : data.incident_expire_timestamp;
+			data.tth = moment.preciseDiff(Date.now(), incidentExpiration * 1000, true)
+			data.distime = moment(incidentExpiration * 1000).tz(geoTz(data.latitude, data.longitude).toString()).format(config.locale.time)
+			
+			// Stop handling if it already disappeared or is about to go away
+			if (data.tth.firstDateWasLater || ((data.tth.hours * 3600) + (data.tth.minutes * 60) + data.tth.seconds) < minTth) {
+				log.warn(`${data.name} Incident already disappeared or is about to go away in: ${data.tth.hours}:${data.tth.minutes}:${data.tth.seconds}`)
+				resolve([])
+				return null
+			}
```
Kneckter: We are not passing the TTH so we should not need this check. We can do that NOW() check in the other file like your TODO had.

Kneckter:
Besides the above comments, I tested this with RDM and MAD formats manually and it worked for both.

src/controllers/incident.js
fosJoddie: tth is calculated a few lines above, so this code should prevent any late discoveries, just like with pokemons.

fosJoddie:
Fixed the issues you mentioned @Kneckter , thanks for reporting. 

I gotta setup my own test instance and look for that error. Does it happen to you all the time while testing or only sometimes?

Kneckter:
@fosJoddie the error looks like it was just a me-issue related to the limit on inotify. https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers#the-technical-details

I did not get it after in creasing the limit. I tested your updates and did not have an issue using manual data

KartulUdus:
`There's no way to see what incidents you track with !tracked - it has to be added in later.`

I don't see a point to add it to !tracked unless monster is in the tracking

Kneckter:
The `!incident remove` command is not removing DB entries

Kneckter:
I tested out the `!invasion` and `!invasion remove` commands and they worked. This was with live RDM data.

ttanguay:
The gruntType field needs to be NOT NULL since it's included in a PRIMARY KEY... in the db migration5 file.

PR #72, opened 2019-10-14
Some MAD specific changes for boosted weather and ex eligible gyms/raids.

t00fan:
MAD passes data a little different for boosted mons and ex eligible gyms/raids.  I have been using this and it's been working fine, but I'm on an older branch.  I am assuming this will work just fine with the latest develop changes.  If this has been fixed in a different way in the current code, please let me know.


tokar86a:
So is this going to be merge?

PR #77, opened 2019-10-30
Replace mustache prefix placeholder in telegram help message

IchHabRecht:
This patch replaces the existing {{prefix}} placeholder in telegram help message. This is done in the original discord help handling as well.

### DIFF
--- a/src/helpers/telegram/commands/help.js
+++ b/src/helpers/telegram/commands/help.js
@@ -31,7 +31,7 @@ module.exports = (ctx) => {
 				greeting.forEach((field) => {
 					message = message.concat(`\n\n${field.name}\n\n${field.value}`)
 				})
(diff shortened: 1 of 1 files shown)


KartulUdus:
don't replace it with config.discord.prefix 

telegram prefixes are always / so 
`message.replace(/{{prefix}}/g, '/'`

IchHabRecht:
@KartulUdus 

Thank you for your very quick reply. Would you mind to assist what needs to be done for getting all tests running?

KartulUdus:
no way you affected that 
but I'm confused how it first passed on master and now on rerun fails 😐 

pls hold while I figure this out before mergning

IchHabRecht:
Would you mind to restart again. It seems some external service may was down. I pushed another branch to my fork and tests run again

KartulUdus:
usually i'd ask to bump the version in package.json, but as new major version is close, it ok

PR #89, opened 2019-12-08
{end} for egg alarms

ccev:
Have the Raid endtime in Egg alarms. Added this a few months back locally and couldn't live without.
Only tested for MAD


jfberry:
Would like to  accept, can you rebase PR?

PR #109, opened 2020-01-22
Fix the Telegram restore feature

gcm-explo1t:
##  Description
Restoring more then 1 Pokemon with the restore feature in telegram did not work, as only a single instance of {{ target }} would be replaced. Thus all the other entries in the DB still have {{ target }} set as ID. With the new search string, every instance of {{ target }} is replaced with the user id.

## How Has This Been Tested?
Yes.

## Types of changes

## Checklist:

### DIFF
--- a/src/helpers/telegram/commands/restore.js
+++ b/src/helpers/telegram/commands/restore.js
@@ -57,7 +57,7 @@ module.exports = (ctx) => {
 							controller.query.deleteQuery('quest', 'id', target.id).catch((O_o) => {}),
 							controller.query.deleteQuery('incident', 'id', target.id).catch((O_o) => {}),
 						]).then((x) => {
-							const query = fs.readFileSync(path.join(__dirname, '../../commando/commands', `/filterBackups/${args[0]}.sql`), 'utf8').replace('{{ target }}', target.id)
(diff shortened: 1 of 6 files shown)


src/helpers/telegram/commands/restore.js
```diff
@@ -57,7 +57,7 @@ module.exports = (ctx) => {
 							controller.query.deleteQuery('quest', 'id', target.id).catch((O_o) => {}),
 							controller.query.deleteQuery('incident', 'id', target.id).catch((O_o) => {}),
 						]).then((x) => {
-							const query = fs.readFileSync(path.join(__dirname, '../../commando/commands', `/filterBackups/${args[0]}.sql`), 'utf8').replace('{{ target }}', target.id)
+							const query = fs.readFileSync(path.join(__dirname, '../../commando/commands', `/filterBackups/${args[0]}.sql`), 'utf8').replace(/'{{ target }}'/g, target.id)
```
KartulUdus: I think quotes would break this 

`/{{ target }}/g`  pls

KartulUdus: also pr's base to develop pls

gcm-explo1t: I only tested with quotes.

KartulUdus: it works there because you're replacing the string in the query with a number. 

for other Id's like webhooks where the id is a varchar it would error as those aren't numbers, but strings

PR #136, opened 2020-05-10
update the wiki link

quiddity-wp:
just updating the link


quiddity-wp:
No longer needed, update was pulled from develop

PR #156, opened 2020-07-03
Update util.json - Adds move sets 333-341 (referenced from pmsf)

kbtbc:
Adds move sets 333-341 (referenced from pmsf)

Not tested


kbtbc:
opened new against develop

PR #173, opened 2020-09-02
Mega Energy Quest Tracking for v3

Kneckter:
## Description
This update adds quest tracking for mega energy to the v3 branch. The commands are `!quest energy1` and `!quest remove energy`. 

You'll need an updated parser that includes the correct reward info (RDM has an open PR for it).

## How Has This Been Tested?
Tested it with my setup today since the new quests were out.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)


src/controllers/quest.js
```diff
@@ -343,6 +348,16 @@ class Quest extends Controller {
 						conditionString = conditionString.concat(cond)
 						break
 					}
+					case 37: {
+						const pokemons = []
+						condition.info.raid_pokemon_evolutions.forEach((pokemonId) => {
+							pokemons.push(monsterData[pokemonId].name)
+						})
+						const template = this.qdts.questConditions['37']
```
IchHabRecht, line 356: This condition isn't defined anywhere and breaks current systems.

Kneckter: Can you paste the error you're getting? What you say doesn't make sense to me because `case 37` is in case the parser sends a quest with the condition of 37. The line `const template = this.qdts.questConditions['37']` looks in your questdts.json file for the string template for forming the output text. I did forget to commit the example update so I pushed that but I would like to see your error since you can start the bot without that dts.

IchHabRecht: There is an error when trying to access qdts.questConditions['37'] as the property doesn't exits. Thanks for updating the example json file.

Kneckter: Let me know if updating the questdts solved that issue. If not, give me more info on what you were doing because poraclejs is running with this code. I haven't scanned a quest with condition 37 yet (I don't think it's been released)

jfberry:
Sorry, no more v3 pull requests will be merged. Please come to discord if there is a reason you can't move to v4 now we have telegram support and we will try and help

PR #184, opened 2020-09-28
Rollup of fixes - eggs, invasion, sqlite compatibility

jfberry:
## Description

Several changes previously described on discord repository:
* Egg notification only to single user (issue 171)
* Quest distance notification not working for sqlite users
* Invasion notification only works for everything
* Sqlite does not like = true
* Bracketing in sql expressions causes unintended expression evaluation
* Tidy up: Unneeded group by clauses (not doing any aggregates)

## How Has This Been Tested?
Tested in production on live instance, various users given fixes of egg and invasion fix
Not tested against mysql but see no reason it will not work

## Types of changes

## Checklist:


jfberry:
Added duplicate in error

PR #185, opened 2020-09-28
Fixes for: sqllite compatibility, raid egg notification, invasion tra…

jfberry:
## Description

Several changes previously described on discord repository:
* Egg notification only to single user (issue 171)
* Quest distance notification not working for sqlite users
* Invasion notification only works for everything
* Sqlite does not like = true
* Bracketing in sql expressions causes unintended expression evaluation

## How Has This Been Tested?
Tested in production on live instance, various users given fixes of egg and invasion fix
Not tested against mysql but see no reason it will not work

## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -48,7 +48,7 @@ class Monster extends Controller {
 		} else {
 			query = query.concat(`
 				and ((monsters.distance = 0 and (${areastring})) or monsters.distance > 0)
-				group by humans.id, humans.name, monsters.template 
+			   group by humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping
 			`)
 		}
 		let result = await this.db.raw(query)
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -43,18 +43,16 @@ class Monster extends Controller {
 			  + sin( radians(${data.latitude}) )
 			  * sin( radians( humans.latitude ) ) ) < monsters.distance and monsters.distance != 0) or
 			   monsters.distance = 0 and (${areastring}))
-			   group by humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping
 			`)
 		} else {
 			query = query.concat(`
-				and (monsters.distance = 0 and (${areastring}) or monsters.distance > 0)
-				group by humans.id, humans.name, monsters.template 
+				and ((monsters.distance = 0 and (${areastring})) or monsters.distance > 0)
 			`)
 		}
 		let result = await this.db.raw(query)
 
 		if (!['pg', 'mysql'].includes(this.config.database.client)) {
-			result = result.filter((res) => res.distance === 0 || +res.distance > 0 && +res.distance > this.getDistance({ lat: res.latitude, lon: res.longitude }, { lat: data.latitude, lon: data.longitude }))
+			result = result.filter((res) => +res.distance === 0 || +res.distance > 0 && +res.distance > this.getDistance({ lat: res.latitude, lon: res.longitude }, { lat: data.latitude, lon: data.longitude }))
 		}
 		result = this.returnByDatabaseType(result)
 		// remove any duplicates
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -43,10 +43,12 @@ class Monster extends Controller {
 			  + sin( radians(${data.latitude}) )
 			  * sin( radians( humans.latitude ) ) ) < monsters.distance and monsters.distance != 0) or
 			   monsters.distance = 0 and (${areastring}))
+			   group by humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping
 			`)
 		} else {
 			query = query.concat(`
 				and ((monsters.distance = 0 and (${areastring})) or monsters.distance > 0)
+				group by humans.id, humans.name, monsters.template 
 			`)
 		}
 		let result = await this.db.raw(query)
--- a/src/controllers/pokestop.js
+++ b/src/controllers/pokestop.js
@@ -32,7 +32,7 @@ class Pokestop extends Controller {
 		} else {
 			query = query.concat(`
 				and ((invasion.distance = 0 and (${areastring})) or invasion.distance > 0)
-				group by humans.id, humans.name, invasion.template
+				group by humans.id, humans.name, humans.type, humans.latitude, humans.longitude, invasion.template, invasion.distance, invasion.clean, invasion.ping
 			`)
 		}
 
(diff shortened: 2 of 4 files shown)


jfberry:
Reverted change to group by after discussion on discord

src/controllers/monster.js
```diff
@@ -47,14 +47,14 @@ class Monster extends Controller {
 			`)
 		} else {
 			query = query.concat(`
-				and (monsters.distance = 0 and (${areastring}) or monsters.distance > 0)
+				and ((monsters.distance = 0 and (${areastring})) or monsters.distance > 0)
```
KartulUdus, line 50: just extra parentheses 

(2 + 2) is the same as ((2 + 2))

src/controllers/monster.js
```diff
@@ -47,14 +47,14 @@ class Monster extends Controller {
...
 				group by humans.id, humans.name, monsters.template 
 			`)
 		}
 		let result = await this.db.raw(query)
 
 		if (!['pg', 'mysql'].includes(this.config.database.client)) {
-			result = result.filter((res) => res.distance === 0 || +res.distance > 0 && +res.distance > this.getDistance({ lat: res.latitude, lon: res.longitude }, { lat: data.latitude, lon: data.longitude }))
+			result = result.filter((res) => +res.distance === 0 || +res.distance > 0 && +res.distance > this.getDistance({ lat: res.latitude, lon: res.longitude }, { lat: data.latitude, lon: data.longitude }))
```
KartulUdus, line 57: result will always be an int, but calling toInt on it doesn't hurt

src/controllers/pokestop.js
```diff
@@ -1,6 +1,7 @@
 const geoTz = require('geo-tz')
 const moment = require('moment-timezone')
 const Controller = require('./controller')
+const { log } = require('../lib/logger')
```
KartulUdus, line 4: you're requiring logger, but not using it

src/controllers/pokestop.js
```diff
@@ -13,8 +14,8 @@ class Pokestop extends Controller {
 		let query = `
 		select humans.id, humans.name, humans.type, humans.latitude, humans.longitude, invasion.template, invasion.distance, invasion.clean, invasion.ping from invasion
 		join humans on humans.id = invasion.id
-		where humans.enabled = true and
-		(invasion.grunt_type='${data.gruntType}' or invasion.grunt_type = 'everything') and
+		where humans.enabled = 1 and
```
KartulUdus, line 17: true and 1 are the same

src/controllers/pokestop.js
```diff
@@ -13,8 +14,8 @@ class Pokestop extends Controller {
...
+		(invasion.grunt_type='${data.gruntType.toLowerCase()}' or invasion.grunt_type = 'everything') and
```
KartulUdus, line 18: hmm, did you find gruntType not matching due to capitalization?

src/controllers/pokestop.js
```diff
@@ -30,7 +31,7 @@ class Pokestop extends Controller {
 			`)
 		} else {
 			query = query.concat(`
-				and (invasion.distance = 0 and (${areastring}) or invasion.distance > 0)
+				and ((invasion.distance = 0 and (${areastring})) or invasion.distance > 0)
```
KartulUdus, line 34: just extra parenthesis like above

src/controllers/quest.js
```diff
@@ -11,10 +11,10 @@ class Quest extends Controller {
 			areastring = areastring.concat(`or humans.area like '%"${area}"%' `)
 		})
 		let query = `
-		select humans.id, humans.name, humans.type, quest.distance, quest.clean, quest.ping, quest.template from quest
+		select humans.id, humans.name, humans.type, humans.latitude, humans.longitude, quest.distance, quest.clean, quest.ping, quest.template from quest
 		join humans on humans.id = quest.id
 		where humans.enabled = 1 and
-		((reward_type=7 and reward in (${data.rewardData.monsters}) and shiny = 1 and ${data.isShiny}=1 or reward_type=7 and reward in (${data.rewardData.monsters}) and shiny = 0)
+		(((reward_type=7 and reward in (${data.rewardData.monsters}) and shiny = 1 and ${data.isShiny}=1) or (reward_type=7 and reward in (${data.rewardData.monsters}) and shiny = 0))
```
KartulUdus, line 17: doesn't look like extra, good. What does this fix exactly?

src/controllers/raid.js
```diff
@@ -32,16 +32,18 @@ class Raid extends Controller {
 			`)
 		} else {
 			query = query.concat(`
-				and (raid.distance = 0 and (${areastring}) or raid.distance > 0)
+				and ((raid.distance = 0 and (${areastring})) or raid.distance > 0)
```
KartulUdus, line 35: just extra parenthesis

src/controllers/raid.js
```diff
@@ -360,8 +363,8 @@ class Raid extends Controller {
 					jobs.push(work)
 					this.addDiscordCache(cares.id)
 				}
-				return jobs
 			}
+			return jobs
```
KartulUdus, line 367: looks like the most important line in the pr. 

was raid not returning all jobs?

KartulUdus (changes requested):

src/controllers/pokestop.js
jfberry: Problems with sqlite using true as this is translated to an int field in the database rather than a boolean

src/controllers/pokestop.js
jfberry: Yes, exactly that (sqlite)

src/controllers/quest.js
jfberry: This is an operator prescience issue with and/or - found sqllite (at least) getting confused with out the explicit use of brackets.

src/controllers/raid.js
jfberry: Correct only first egg returned (issue created)

src/controllers/monster.js
jfberry: There were other operator prescience related problems (eg putting.  1 and 2 or 3 when what was meant was ((1 and 2) or 3) - some of the ones I edited may have been not required either for correctness or to be explicit - I will review each of them

src/controllers/pokestop.js
jfberry: Remains from debugging, will remove

src/controllers/pokestop.js
jfberry: see other comment

src/controllers/raid.js
jfberry: see other comment

dkmur:
just tried ``quest spinda distance1000`` and it does not take the distance part, remains 0m so using area. On mariadb, could be related to this PR?

dkmur:
> just tried `quest spinda distance1000` and it does not take the distance part, remains 0m so using area. On mariadb, could be related to this PR?

bloody idiot use d1000

jfberry:
Previously returned group by (as requested) caused missing messages because some needed rows were not returned. Have fixed the group by statements rather than reverting to previous situation of not having them

src/controllers/quest.js
jfberry: Also missing latitude and longitude means that distance based notification not working

PR #188, opened 2020-10-29
Resolve missing messages user xxx not found

jfberry:
## Description

When Poracle is restarted some users are not in the local discord.js cache meaning that any direct messages to them fail. This simple fix detects that scenario and creates a conversation with them

## How Has This Been Tested?

Multiple users on discord have reported this successfully resolves their problems

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/discord/discordWorker.js
+++ b/src/lib/discord/discordWorker.js
@@ -77,10 +77,14 @@ class Worker {
 	}
 
 	async userAlert(data) {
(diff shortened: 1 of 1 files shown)


cks85:
Working as intended, can see in logs that it is recreating the DM when not interaction from the user has been detected since restart, and notifications are going through to the user without any action from him (wasn't the case before this change).

PR #191, opened 2020-10-29
Special case tracking pokemon 'everything' instead of having individual entries for each mon

jfberry:
## Description
The existing implementation for !track everything adds a database entry for every single pokemon. This is inefficient and results in a massive !tracked report. This change allows a '0' to be entered under the pokemon id representing all pokemon.
It does not require a database change as the old format still works. If a user does !untrack everything it will still remove the old entries so the database will gradually move to the new format

## How Has This Been Tested?

Tested locally, would appreciate feedback

## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -16,7 +16,7 @@ class Monster extends Controller {
 		select humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping from monsters
 		join humans on humans.id = monsters.id
 		where humans.enabled = true and
-		pokemon_id=${data.pokemon_id} and
+		(pokemon_id=${data.pokemon_id} or pokemon_id=0) and
 		min_iv<=${data.iv} and
 		max_iv>=${data.iv} and
 		min_cp<=${data.cp} and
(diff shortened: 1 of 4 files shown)


Piikachuu:
Hi thank for this pr , but i have issue with this :/ 
when i track everyhting on log i dont have list of mon tracked and on db i have only 1 row empty so nothing is sended..

log poracle:
image

db:
image

!tracked result
image

jfberry:
The point is that everything is now a single row in the database with pokemon_id = 0
You can see that the query change in monster.js will now pick this and match against all pokemon.
So it should be the case that there is one row but the alerts should still function.

Have you found an alert that you have missed?

Piikachuu:
its ok all is fine after multi tests ;)

jfberry:
Updated to special case 'gen' - eg !track everything gen1 will put the individual mon in the database rather than falling back to the everything special case

kamieniarz:
This PR increased my CPU usage by a lot so sth is wrong with it in my case

jfberry:
Interesting. If you are sure it is this PR then hit me up on discord and we can work through it - the only thing it could be is the modification of the query (to check for =0 as well as a mon). Are you mysql or sqlite? Is mysql on the same box? Is it mysql which is hogging the CPU?

PR #192, opened 2020-10-29
Reduce memory use for duplicate check cache

jfberry:
## Description
Modify the use of the duplicate message cache to reduce memory

## Motivation and Context

Some users have very large memory footprints, the storage of the whole message to eliminate duplicates is not required.

## Types of changes

## Checklist:

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -23,7 +23,7 @@ const readDir = util.promisify(fs.readdir)
 const telegraf = new Telegraf(config.telegram.token, { channelMode: true })
 
 
-const cache = new NodeCache({ stdTTL: 5400 })
+const cache = new NodeCache({ stdTTL: 5400, useClones: false })
 
 const discordCache = new NodeCache({ stdTTL: config.discord.limitSec })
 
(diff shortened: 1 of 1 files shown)


cks85:
Significant saving of RAM usage by Poracle with these changes. Not creating issues as far as I noticed after 12 hours running.

KartulUdus:
I don't know if `.has`is better to use 

however that 
`JSON.stringify(hook))`

is needed so that it would correctly cache raids

jfberry:
> I don't know if `.has`is better to use
> 
> however that
> `JSON.stringify(hook))`
> 
> is needed so that it would correctly cache raids

Perhaps I'm missing something somewhere else?  I can't see anywhere where the contents of the cache are used (and I checked in the controllers also), it only appears to verify that the key exists to determine whether the raid & pokemon has been seen before.
Two of the incoming web hook types already just used the 'cache' check rather than a stringified version of the hook

has vs get - this is marginal; I checked the source for the cache and this is a shorter codepath if you are not going to use the value, which this code does not appear to.  The core reason for the PR though is to avoid caching the whole hook (removing the JSON.stringify) - as you can imagine especially with pokemon this takes up a considerable amount of memory

PR #199, opened 2020-11-17
Live through nominatim service being unavailable

jfberry:
A reasonably common operational problem is the unavailability of the nominatim server. This falls back to an 'unknown' text if not available

## Description

Simply catch the error

## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/controller.js
+++ b/src/controllers/controller.js
@@ -133,7 +133,7 @@ class Controller {
 			geoCache.save(true)
 			return result
 		} catch (err) {
-			throw { source: 'getAddress', error: err }
+			return { addr: 'Unknown', flag: '' }
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/controllers/controller.js
```diff
@@ -133,7 +133,7 @@ class Controller {
 			geoCache.save(true)
 			return result
 		} catch (err) {
-			throw { source: 'getAddress', error: err }
+			return { addr: 'Unknown', flag: '' }
```
lenisko, line 136: Going through with notification is very good idea but completely silencing error is not.
How about adding this to error log?
`this.log.error('getAddress: failed to fetch data', err)`

jfberry: I was thinking about the permanent 'not going to install nominatim' situation and log bloat - but you're right this should perhaps be through a config.  I will add a log and do something different for a 'none' situation

LucianoNeo: Working like a charm, thank you!

jfberry:
Variant pushed

PR #202, opened 2020-11-18
Add command `!tracked area`

kamieniarz:
## Description
New command is ment to display tracked areas only. I'm not sure if I've done it in a "nice" way but it works tho 🤷

## Motivation and Context
Make listing tracked areas easier

## How Has This Been Tested?
Tested locally - works

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/discord/commando/commands/tracked.js
+++ b/src/lib/discord/commando/commands/tracked.js
@@ -39,6 +39,10 @@ exports.run = async (client, msg, [args]) => {
 		const quests = await client.query.selectAllQuery('quest', { id: target.id })
 		const invasions = await client.query.selectAllQuery('invasion', { id: target.id })
 		const maplink = `https://www.google.com/maps/search/?api=1&query=${human.latitude},${human.longitude}`
+		if (args.includes('area')) {
+			return msg.reply(`You are currently set to receive alarms in ${human.area}`)
+		}
+		if (!args.includes(args[0])) {
 		let locationText = ''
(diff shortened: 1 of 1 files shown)


src/lib/discord/commando/commands/tracked.js
```diff
@@ -39,6 +39,10 @@ exports.run = async (client, msg, [args]) => {
 		const quests = await client.query.selectAllQuery('quest', { id: target.id })
 		const invasions = await client.query.selectAllQuery('invasion', { id: target.id })
 		const maplink = `https://www.google.com/maps/search/?api=1&query=${human.latitude},${human.longitude}`
+		if (args.includes('area')) {
+			return msg.reply(`You are currently set to receive alarms in ${human.area}`)
+		}
+		if (!args.includes(args[0])) {
```
jfberry, line 45: You don't need the second 'if' statement because you are doing a return above.
You could argue that this whole if section could be higher up to avoid the queries.
Should make area translatable:
eg args.includes(client.translator.translate('area'))

jfberry:
If we are to include options on tracked I wonder if the right thing would be to have a series of possible parameters eg
areas location monsters eggs quests raids to see that/those sections?

However I am not entirely sure of the utility of this now that !tracked has been shrunk by the !track everything patch which removes thousands of lines from the output...

jfberry:
I think this has unwound the webhook change
All text in file is now translated

PR #219, opened 2020-12-10
add user pvp preferences for dynamic DTS

plinytheelder:
adds new DTS variables to serve the following purposes:

- removes hardcoded values from current pvp DTS and bases it on user preferences.
- removes unnecessary pvp stats from notifications if a user didn't request them (e.g. above their threshold but below current hardcoded DTS values, or if they didn't set up pvp notifications and don't care about pvp at all).
- allows for 1 pokemon template, instead of multiple for IVs, great league, ultra league, etc.

EXAMPLE DTS:

{{#each pvp_rankings_great_league}}{{#compare this.rank '<=' ../greatleagueranking}}<ADD YOUR DESIRED OUTPUT HERE>{{/compare}}{{/each}}

{{#each pvp_rankings_ultra_league}}{{#compare this.rank '<=' ../ultraleagueranking}}<ADD YOUR DESIRED OUTPUT HERE>{{/compare}}{{/each}}

### DIFF
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -13,7 +13,7 @@ class Monster extends Controller {
 			areastring = areastring.concat(`or humans.area like '%"${area}"%' `)
 		})
 		let query = `
-		select humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping from monsters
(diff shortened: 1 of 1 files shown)


lmb439:
Hm. One concern that I have is that the current pvp tracking is insufficient and must be augmented with manual trackers. (e.g. eevee, where tracking rank > 5 will report the top 5 distributions for all eeveelutions)

To fix this, I have been tracking such iv distributions manually with maxLevel, atk maxAtk, etc.

Perhaps an extra argument for reporting pvp stats could be added, and specifying great/ultra/greatcp/ultracp would behave as if that argument were provided as well?

PR #221, opened 2020-12-11
Global toggles, text type templates, PVP filters and display

Petap0w:
PR fixing / adding many features on Poracle Develop branch:
- Added `config > general > disabledCommands` to completely disable certain commands for users (ex. `"disabledCommands": ["invasion", "quest"]`)
- Added global toggles to avoid processing incoming data that is not used at all, and will hide section in user `!tracked` result:
```
    "general": {
        "disablePokemon": false,
        "disableRaid": false,
        "disableInvasion": false,
        "disablePokestop": false,
        "disableQuest": false,
        "disableWeather": false
    }
```
- Added `config > tracking > disableEverythingTracking` to prevent users from using `everything` in their `!track` commands (admin still can)
- Added `config > tracking > forceEverythingSeparately` to force `!track everything` to register 1 line per pokémon in db
- Added `config > tracking > defaultDistance` and `maxDistance` to set a distance when user omits it and/or set a value too high (`0` to disable feature) 
- Changed template type from int to text so real names can be set and used easily by users (ex. create a dts section with a template having `"id": "pvpalert",` and it can be used in tracking commands with `templatepvpalert`) - this requires a DB migration which may fail if having duplicate entries in monsters table (check Discord help channel if that happens)
- Added config settings for minimum PVP filters and DTS display options. `pvpDisplay...` settings will cap the results shown in user alerts to the value set in config (ex. a candidate with pvp rank # 150 will not show whereas the evolution being rank # 8 will show, or a base form rank # 1 reaching 850cp and useless will not show either). `pvpFilter...` settings will force boundaries on the user tracking requests when they are out of range.
```
    "pvp": {
        "pvpDisplayMaxRank": 10,
        "pvpDisplayGreatMinCP": 1450,
        "pvpDisplayUltraMinCP": 2450,
        "pvpFilterMaxRank": 10,
        "pvpFilterGreatMinCP": 1450,
        "pvpFilterUltraMinCP": 2450
    },
```
- Added configurable path to geofence file in config and also support for GEOjson geofence file format. `defaultGeofenceName` is used in the conversion to Poracle format if missing from the source file combined with a unique id (`defaultGeofenceColor` if someday that is used for something).
```
    "geofence": {
        "path": "./config/geofence.json",
        "defaultGeofenceName": "city",
        "defaultGeofenceColor": "#3399ff"
    },
```
- Added user notice when he tries to register a tracking command with a distance not having set his location, or without a distance and not having set his area
- Fixed some translations in tracking commands and improved the presentation of results in `tracked`
- Global update of `fr.json` locale (and of `de.json` which needs additional contribution)
- Addition of Pikachu Dec. 2020 forms in `monsters.json`

## Types of changes


jfberry:
There is some great stuff in here! A little difficult to unpack all the changes but I think there are a few things to fix up:
* The database migrations are applied (I think) in alphabetical order. So this one would try to run before 'base' in a new install which would fail.  Perhaps we should start numbering them (would have to start with a letter now since we missed the opportunity before -- eg q00whatever, q01whatever, q02whatever...) - q to be after the pvp change.
* Config file changes -- would be best to have defaults for these new options so it can run with an unaltered local.json
* Template change from numeric -- would be best to run with unaltered dts.json, haven't looked at the template matching code wonder if there is a way we can keep existing files working
* Translation work - I intend to take bbdoc's translation patch first as a baseline for further language work because he has gone to the effort of rolling up all previous translation PRs

Petap0w:
> * The database migrations are applied (I think) in alphabetical order. So this one would try to run before 'base' in a new install which would fail.  Perhaps we should start numbering them (would have to start with a letter now since we missed the opportunity before -- eg q00whatever, q01whatever, q02whatever...) - q to be after the pvp change.
-> Fixed with `v00_` prefix
> * Config file changes -- would be best to have defaults for these new options so it can run with an unaltered local.json
-> that is what default.json is for (is missing options from local.json they are taken from default.json)
> * Template change from numeric -- would be best to run with unaltered dts.json, haven't looked at the template matching code wonder if there is a way we can keep existing files working
-> fixed should be compatible now
> * Translation work - I intend to take bbdoc's translation patch first as a baseline for further language work because he has gone to the effort of rolling up all previous translation PRs
-> included all bbdoc's translation and added more

PR #225, opened 2020-12-19
Add option to force use of location or area in tracking

Petap0w:
Add `config > tracking > forceLocationOrArea` to force users having to set a location or area(s) for their tracking commands (default `false`)

## Types of changes


jfberry:
Not sure this option is needed -- I can't understand why you would want to turn this off?

PR #227, opened 2020-12-19
fix need of duplicate capitalized translation

Petap0w:
Fix the need of duplicate translation with different capitalization.


jfberry:
I see what you're doing, but normally when translating you do this separately as they as separate usages (often you deliberately try to individualise strings).  Since you guys are doing all the translation work if you think this is for the best I'll merge it but just think whether this is what you _actually_ want to do...

PR #237, opened 2020-12-29
Rework of energy quest tracking

Petap0w:
Rework of Quest Mega Energy tracking:
- tracking is now on target pokemon's mega energy and not energy amount
- you can `!quest energy` for all pokemons, or `!quest energycharizard energyblastoise` for specific targets
- `!quest everything` will also include all mega energy types

Changes also in `!tracked` command:
Capture d’écran 2020-12-29 à 09 41 18

Changes in dts to hide Conditions when there are none:
Capture d’écran 2020-12-29 à 09 39 39
Capture d’écran 2020-12-29 à 09 40 22


jfberry:
Would be nice to get comments from users who are running, plan to take this PR after some user confirmation

PR #239, opened 2020-12-29
Allow direct tracking of pvp evolved forms

Petap0w:
PVP tracking can now be done directly on evolved form when enabled with new `config` > `pvp` > `pvpEvolutionDirectTracking` setting

`!track charizard great5` will alert for wild scanned charmander which once evolved will become Charizard with rank < 5 (and also for wild charizard if they match pvp requirements).

This is using the `pvp_rankings_great_league` and `pvp_rankings_ultra_league` data included in the scanned pokémons (so it has to be provided by the scanner: RDM or MAD plugin).

Setting tracking commands on the non evolved form available in the wild will still alert on all evolutions that are matching the rank / cp filters in the command:
`!track charmander great5` will alert on charmander / charmeleon / charizard rank < 5 (unless additional filters are set on minCp, in the command or via general config options)

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -45,6 +45,7 @@
         "defaultGeofenceColor": "#3399ff"
     },
     "pvp": {
+        "pvpEvolutionDirectTracking": false,
         "pvpDisplayMaxRank": 10,
         "pvpDisplayGreatMinCP": 1450,
         "pvpDisplayUltraMinCP": 2450,
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -12,8 +12,8 @@ class Monster extends Controller {
 		data.matched.forEach((area) => {
 			areastring = areastring.concat(`or humans.area like '%"${area}"%' `)
 		})
-		let pokemonQueryString = `(pokemon_id=${data.pokemon_id} or pokemon_id=0)`
-		if (data.pvpEvoLookup) pokemonQueryString=`(pokemon_id=${data.pvp_pokemon_id} and (great_league_ranking < 4096 or ultra_league_ranking < 4096 or great_league_ranking_min_cp > 0 or ultra_league_ranking_min_cp > 0))`
+		let pokemonQueryString = `(pokemon_id=${data.pokemon_id} or pokemon_id=0) and (form = 0 or form = ${data.form})`
+		if (data.pvpEvoLookup) pokemonQueryString=`(pokemon_id=${data.pvp_pokemon_id} and (form = 0 or form = ${data.pvp_form}) and (great_league_ranking < 4096 or ultra_league_ranking < 4096 or great_league_ranking_min_cp > 0 or ultra_league_ranking_min_cp > 0))`
 		let query = `
 		select humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping, monsters.great_league_ranking, monsters.ultra_league_ranking from monsters
 		join humans on humans.id = monsters.id
@@ -24,7 +24,6 @@ class Monster extends Controller {
 		min_cp<=${data.cp} and
 		max_cp>=${data.cp} and
 		(gender = ${data.gender} or gender = 0) and
-		(form = 0 or (form = ${data.form} and ${data.pvpEvoLookup}=0) or (form=${data.pvp_form} and ${data.pvpEvoLookup}=1)) and
 		min_level<=${data.pokemon_level} and
 		max_level>=${data.pokemon_level} and
 		atk<=${data.individual_attack} and
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -16,7 +16,7 @@ class Monster extends Controller {
 		select humans.id, humans.name, humans.type, humans.latitude, humans.longitude, monsters.template, monsters.distance, monsters.clean, monsters.ping, monsters.great_league_ranking, monsters.ultra_league_ranking from monsters
 		join humans on humans.id = monsters.id
 		where humans.enabled = true and
-		(pokemon_id=${data.pokemon_id} or pokemon_id=0) and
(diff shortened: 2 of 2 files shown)


plinytheelder:
testing right now. seems to work as intended. only unintended consequence is I appear to be receiving 2 posts for every spawn. anyone else?

Petap0w:
> testing right now. seems to work as intended. only unintended consequence is I appear to be receiving 2 posts for every spawn. anyone else?

new commit pushed which should fix that

valleypogo:
ok. better. but one more thing. something is funky getting pokemon_id to pass to staticmaps. Logs show:

Failed to Pregenerate StaticMap. Got 500. Error: ImageMagick Error: convert: improper image header

plinytheelder:
ok. better. but one more thing. something is funky getting pokemon_id to pass to staticmaps. Logs show:

Failed to Pregenerate StaticMap. Got 500. Error: ImageMagick Error: convert: improper image header

plinytheelder:
when I tried to call tileserver directly using the URL and DTS, logs showed

Cannot read property 'toString' of undefined {"stack":"TypeError: Cannot read property 'toString' of undefined

Petap0w:
> when I tried to call tileserver directly using the URL and DTS, logs showed
> 
> Cannot read property 'toString' of undefined {"stack":"TypeError: Cannot read property 'toString' of undefined

I don't see how this could be related to what this PR does, we aren't removing anything / changing anything existing in `data`... just adding a few more entries in it that aren't used in `getPregeneratedTileURL()`

jfberry:
This seems like the right approach but will be more cautious about merging this until has had some user testing. Default is off, which is correct for now.  I will bring into my local copy and run it to see impact on performance to see if any fine tuning needed, but would like other comments too before merge please!

valleypogo:
> > when I tried to call tileserver directly using the URL and DTS, logs showed
> > Cannot read property 'toString' of undefined {"stack":"TypeError: Cannot read property 'toString' of undefined
> 
> I don't see how this could be related to what this PR does, we aren't removing anything / changing anything existing in `data`... just adding a few more entries in it that aren't used in `getPregeneratedTileURL()`

ok. I see the issue. it isn't a problem with pokemon_id. it's form id. I used test DTS and noticed that {{pokemon_id}} was correct and returning the ID of the base mon, but {{form}} was returning the form id of the evolved mon. So that was what was causing static maps to fail.

plinytheelder:
> > when I tried to call tileserver directly using the URL and DTS, logs showed
> > Cannot read property 'toString' of undefined {"stack":"TypeError: Cannot read property 'toString' of undefined
> 
> I don't see how this could be related to what this PR does, we aren't removing anything / changing anything existing in `data`... just adding a few more entries in it that aren't used in `getPregeneratedTileURL()`

ok. I see the issue. it isn't a problem with pokemon_id. it's form id. I used test DTS and noticed that {{pokemon_id}} was correct and returning the ID of the base mon, but {{form}} was returning the form id of the evolved mon. So that was what was causing static maps to fail.

Petap0w:
> > > when I tried to call tileserver directly using the URL and DTS, logs showed
> > > Cannot read property 'toString' of undefined {"stack":"TypeError: Cannot read property 'toString' of undefined
> > 
> > 
> > I don't see how this could be related to what this PR does, we aren't removing anything / changing anything existing in `data`... just adding a few more entries in it that aren't used in `getPregeneratedTileURL()`
> 
> ok. I see the issue. it isn't a problem with pokemon_id. it's form id. I used test DTS and noticed that {{pokemon_id}} was correct and returning the ID of the base mon, but {{form}} was returning the form id of the evolved mon. So that was what was causing static maps to fail.

should now be fixed

plinytheelder:
tried again and see the same issue: {{form}} is returning the evolved mon which is in turn breaking static map images and said DTS. easily noticed on mon with shadow forms. most aren't an issue since 0 is static for both.

weird though, form is correct within the {{imageUrl}} and the sprite icon is displaying fine there.

Petap0w:
> tried again and see the same issue: {{form}} is returning the evolved mon which is in turn breaking static map images and said DTS. easily noticed on mon with shadow forms. most aren't an issue since 0 is static for both.
> 
> weird though, form is correct within the {{imageUrl}} and the sprite icon is displaying fine there.

might be a cache problem or so as really now no `data` values are changed from initial mon (besides bestrank and bestrankcp)

plinytheelder:
I thought it might be tileserver cache as well. but when I added {{form}} to my DTS, it was still incorrect.

example: Snover returning form 935, when it should be 932.

Petap0w:
> I thought it might be tileserver cache as well. but when I added {{form}} to my DTS, it was still incorrect.
> 
> example: Snover returning form 935, when it should be 932.

sorry, short night, did a `git pull` instead of `git push` for that last commit...

plinytheelder:
no worries. same here. couldn't sleep so figure I'd test this.

anyway, now something seems broken. nothing is triggering and getting 0 person has cared across the board.

Petap0w:
> no worries. same here. couldn't sleep so figure I'd test this.
> 
> anyway, now something seems broken. nothing is triggering and getting 0 person has cared across the board.

let's try once more :D fixed a missing case in the sql query

plinytheelder:
working flawlessly now. good job and a HUGE improvement.

only test left would be CPU monitoring. I do not see a difference, however, the majority of my users rely on a different bot. not many have set them up in poracle...yet. hopefully someone with a good deal of users with pvp notifications could test and report back.

jfberry:
The way it is structured now I think that performance impact will be limited, but the timings are in the log file now (for the sql query).  You can try it off and on and see for yourself what difference it makes on your system. My plan is to try and find a high volume user to test it to see there are no big impacts before merging

plinytheelder:
working well for me. I say merge.

jfberry:
I have now been running for a while, without any issues and sql timings are not significantly impacted. petap0w are you ready for merge?

PR #241, opened 2021-01-01
log all DMs to logchannel

Petap0w:
Based on pr240 from JabLuszko and with addition of auto deletion after some time.

Fill in `config` > `discord` > `dmLogChannelID` with the desired log channel's ID
Optional : `config` > `discord` > `dmLogChannelDeletionTime` : time in minutes after which the logged messages are deleted

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -76,7 +76,9 @@
             "#0070DD",
             "#A335EE",
             "#FF8000"
-        ]
+        ],
+        "dmLogChannelID": "",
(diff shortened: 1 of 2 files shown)


kamieniarz:
Would be great to make it works with PoracleWeb also

jfberry:
> Would be great to make it work with PoracleWeb also

Nothing to stop PoracleWeb posting to the same channel if configured. Because of the way it works though (just changing the tracking database), Poracle would have no idea a change had been made

PR #243, opened 2021-01-03
Add option to watch for config file changes

Petap0w:
⚠️ Totally experimental ⚠️

Seems to be working as intended : modification of `dts.json`, `local.json`, `geofence.json` would be taken without restarting Poracle (any file in `config` dir).
Although there might be other consequences that my limited knowledge of js does not see yet..

Needs a npm package to work: `npm install chokidar` in PoracleJS dir.
Then set new `config` > `general` > `listenFileChanges` to `true`

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -7,6 +7,7 @@
     },
     "general": {
         "environment": "production",
+        "listenFileChanges": false,
         "alertMinimumTime" : 120,
         "imgUrl": "https://raw.githubusercontent.com/nileplumb/PkmnShuffleMap/master/PMSF_icons_large/",
         "locale": "en",
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "homepage": "https://github.com/KartulUdus/reference#readme",
   "dependencies": {
     "axios": "^0.19.0",
+    "chokidar": "^3.4.3",
     "commander": "^4.1.0",
     "config": "^3.2.4",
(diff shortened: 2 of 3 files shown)


t00fan:
This works for monitoring changes and picks up the new changes. However it seems to make the bot create duplicate responses to all commands.  Every time it reloads (because of a file change), the commando bot increments it's response.

For e.g. if I changed the file 5 times and it reloaded 5 times, then when I type !tracked, I get 5 responses from the bot. Number of duplicate responses go up as reloading goes up.  Goes up to the point where you could get temp banned from discord for overusing the API.

jfberry:
How often does poracle get restarted except for an upgrade? I'm experimenting with writing out the messages due to be cleaned on shutdown and bringing them back to life on startup.  That would mean a restart wasn't as painful, would this effectively remove the need for this change?

t00fan:
Yes. If the duplicate notifications and clean problem were solved upon restart, then this change would not be needed.  The primary reason (for me at least) for this request is that on restart (this could be adding admin id or messing with dts.. mainly working with dts) there are duplicate notifications sent.

plinytheelder:
I would also like this for playing with DTS changes.

side note, I was just playing with DTS for like 30 minutes, trying to figure out why none of my changes were picking up correctly...then I realized I kept restarting the wrong pm2 process...

PR #245, opened 2021-01-03
Add option to check periodically for member role

Petap0w:
Add the option to set an interval for member role verification: will compare users in Poracle dB to discord server and remove users without the valid user role.

Set the guild ids in `config` > `discord` > `guilds`
Toggle on the feature `config` > `discord` > `checkRole`
Set the verification interval (in hours) `config` > `discord` > `checkRoleInterval`

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -65,7 +65,7 @@
         "unknownResponse": false,
         "invite": "",
         "token": [],
-        "guild": "",
+        "guilds": ["358180805031493634"],
         "channels": ["533927791096233984"],
         "userRole": ["roleId1", "358207117951369217"],
         "admins": ["222742859059560458"],
--- a/config/default.json
+++ b/config/default.json
@@ -60,9 +60,12 @@
     },
     "discord": {
         "enabled": true,
+        "checkRole": false,
+        "checkRoleInterval": 6,
         "unknownResponse": false,
         "invite": "",
         "token": [],
+        "guild": "",
         "channels": ["533927791096233984"],
         "userRole": ["roleId1", "358207117951369217"],
         "admins": ["222742859059560458"],
--- a/src/app.js
+++ b/src/app.js
@@ -120,10 +120,12 @@ async function run() {
 		let roleWorker = new DiscordWorker(config.discord.token[0], 0, config)
 		setInterval(async () => {
 			try {
-				log.info(`Verification of Poracle user's roles starting`)
-				const allUsers = await fastify.monsterController.selectAllQuery('humans', { type: 'discord:user' })
+				let usersToCheck = await fastify.monsterController.selectAllQuery('humans', { type: 'discord:user' })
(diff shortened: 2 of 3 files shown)


jfberry:
Don't think you need a new role id - this is what userRoles is for  -- check out src/lib/discord/commando/events/guidMemberUpdate - it is checking for membership of one of these roles and registering and unregistering accordingly

jfberry:
Not sure you can just safely use one of the discordWorkers without causing a problem. Even though you are checking that its busy you are not setting busy. It seems that syncing it up with an existing worker may be a problem

Perhaps you should use put t whole he membership check in its own file - bring in Discord.js; create a new connection - do the check - and close.  It can then be called on the schedule from the setInterval as you are.

Petap0w:
> Don't think you need a new role id - this is what userRoles is for -- check out src/lib/discord/commando/events/guidMemberUpdate - it is checking for membership of one of these roles and registering and unregistering accordingly

not adding any new role id here

jfberry:
> > Don't think you need a new role id - this is what userRoles is for -- check out src/lib/discord/commando/events/guidMemberUpdate - it is checking for membership of one of these roles and registering and unregistering accordingly
> 
> not adding any new role id here

New configuration for guid in config/discord/guid when this configuration is already in config/discord/userRoles?

Pogodenhelder:
i checked this PR and it works great.
I added a user automaticly by assigning the role for poracle.
Then stopped poracle and delete the role so poracle was not aware of it since it was not running.
Then started poracle again.

It checks the users at a time set in the config, and the user was found to have no assigned role and it was deleted.

Schermafbeelding 2021-01-04 om 13 36 11

only suggestion i would have is to have this function also on startup of the bot before any connection is open from rdm/mad.
Why?
Well if you have in config 24 hours, so it checks every 24 hours and poracle reboots multiple times a day this function is never called and dummy accounts never gets deleted.
For the rest it works as intended.

jfberry:
Poracle should not be rebooting many times a day any longer, please come to discord if it is not stable for you (good suggestion anyway)

Pogodenhelder:
i dont know how many times it reboots it runs on pm2 for months and months, and i think its now on 250 reboots, that also includes my own reboots when im doing things on poracle and when its needed to reboot.

Petap0w:
Pushed new commit with role check 10 sec after startup before going for the config interval.

jfberry:
This is looking good - logs will tell us if we see any problems re-using the discord token while its running.
Only question is whether it would be valuable to have a mode where it runs but only outputs the users it _would_ delete rather than actually doing them?  I can see first use being a little be scary :-)

Pogodenhelder:
indeed if people just bluntly enable it and dont have all their poracle roles in the config it WILL delete the one NOT in the array of the config.. true.
On the other hand its false by default... so if people enable it well... backup first 👍

PR #246, opened 2021-01-04
[Should be feature complete] Telegram support for v4

jfberry:
Well, you asked for it.  Here it is!  Initial telegram support for v4.  It should be feature complete now, and example dts provider!  I plan on leaving as a PR for a little while just in case there are minor niggles, but I have no known issues right now.

Telegram works for monster, raids, quests, eggs, and  invasions notifications; the following commands are supported
```
/poracle
/track
/tracked
/location
/untrack
/start
/stop
/area
/egg
/raid
/invasion
/quest
/channel add (for groups)
/identify (tells you your telegram id and current group id)
/backup
/restore
```

Your dts requires a different format for telegram, here is my test - note you can specify a sticker and whether the location tile is sent. You will need to add your own dts for telegram for each of the message types you are using and examples are in the defaults/dts.json file

``` {
    "id": 1,
    "type": "monster",
    "default": true,
    "platform": "telegram",
    "template": {
      "content": "{{round iv}}% {{name}} cp:{{cp}} L:{{level}} {{atk}}/{{def}}/{{sta}}",
      "sticker": "{{{stickerUrl}}}",
      "location": true
    }
  },
```

The dts can contain a sticker, photo, and location. You can also put webpage_preview: true to get webpage preview turned on in your message.  There is a default sticker repository of webp files and a new stickerUrl expansion

Note that the design allows discord and telegram commands to share the same code.  So Discord commands also need testing as these follow the common code now.  

Full support for channels using channel add with a name (channel add also works in a group)

If you have never set up telegram before, see the wiki - you need to speak to the botfather to get a bot identification

Messages can be 'cleaned' like in discord, so deleted after they become out of date

The configuration file has admins (user ids), and channels (these are the groups that you can do /poracle in to register). You can find these out by sending /identity to poracle and it will tell you the channel and current user id.  Note you should understand the difference between groups and supergroups - supergroups start with -100 and that's probably what you want, google will tell you how to force your group into a supergroup.  Telegram can change a group into a supergroup when it wants and of course the id changes - so that will mean that your alerts and configuration won't match. If you are a v3 user already you will know all about this :-)

Basic guide for getting started:
* Register your bot with the telegram botfather, put the token into poracle configuration
* Now you need to find your user id for admin - send /identify directly to the bot. Put this id as an admin in config. Restart poracle.
* Create a group (not a channel) and invite your bot into the channel
* /identify in the group to get the group id - put that in the channel list in local.json - this is the group where users can do /poracle to get registered. Restart poracle.
* Users can now get registered with poracle using /poracle (you can try this for yourself).
* You can now create groups and once poracle has joined them do **/channel add** to then issue tracking commands for the group
* You can create a channel and have poracle join that, do /identify in there and get the id for the channel. This can be added to poracle in the same way as webhooks (send a direct message with /channel add nameyourname channelid)

TODO:
* npm run init doesn't complete the telegram section


philcerf:
Does it support mutlipled commands per single message (one per line)?

jfberry:
> Does it support mutiple commands per single message (one per line)?

Can you give me an example and I can test?

philcerf:
Well maybe something like:
```
/track Bulbasaur iv100
/track Charmander iv100
/track Squirtle iv100 d2500
```
but sent as one message (with newlines)

jfberry:
Join discord to discuss not sure I understand the benefit of doing this as one line (but it could be done of course)

—
James Berry, m. 07768 021031


________________________________

jfberry:
There was an old implementation (which didn't work) with pipe delimiters which has been stripped out of this (since it didn't work properly) - however, in stripping it out I am actually close to fixing it. Please join discord and I can understand what you are trying to do and find the best solution

philcerf:
@jfberry @KartulUdus .... as @gcm-explo1t mentoned over in #253 there was now some testing of this in our local site (which is quite big, I'd say).... and it seems to work so far.

Can this PR merged?

And perhaps even better... can all the stuff from branch develop be merged into master?

jfberry:
We are planning to merge develop to master within the next week, and then this will be merged into develop. It touches quite a lot of Poracle so it does need really good testing from telegram and discord users hence caution. But lots of people are using it successfully so very confident in it!

gcm-explo1t:
When the fixes @jfberry and me tested are commited, this branch looks good.

jfberry:
The fixes were not strictly related to telegram, but they were manifest for you during telegram testing so have included them in this PR to save me a nightmare merging.
If you could update to be running on this rather than the copy we modified locally and confirm it is good then I am back to no known issues

PR #251, opened 2021-01-10
Fixes issue on PVP display in alerts

Petap0w:
Fixes the issue when enabling pvpEvo direct tracking and a candidate would have two distinct evo matching filters on great / ultra league : one would not keep the bestRank info.


kbtbc:
Tested this, and issue appears to be fixed.  Thank you.

PR #252, opened 2021-01-11
correct german monster names

philcerf:
Signed-off-by: Philippe Cerfon <philcerf@gmail.com>


philcerf:
FYI: From my side, we could close this without merging, since we've now migrated to the devel branch… Though it might be interesting for others still on v3.

Issue #265, opened 2021-01-18
how to upgrade monsters.json and locale/*


philcerf:
Hey.

I've just wanted to do the same what I did for V3 in #252 and #253, i.e. add missing Pokémon and forms and correct typos.

Now first... comparing the pkm names in monsters.json with the list of current official English names:
[list.en.txt](https://github.com/KartulUdus/PoracleJS/files/5827252/list.en.txt)

gives a number of unexpected differences (e.g. use:
`meld  <( grep '^           "name": "' monsters.json | sed 's/^             "name": "//; s/",$//' | uniq ) list.en.txt`
).

Including:
- `Nidoran female` vs `Nidoran♀`
- `Nidoran male` vs ``
- `` vs `Nidoran♂`
- There's a `Nidoran` which maps to the same ID than `Nidoran male`
- `Farfetchd` vs `Farfetch’d`
- `Mr mime` vs `Mr. Mime`
- `Ho oh` vs `Ho-Oh`
- `Mime jr` vs `Mime Jr.`
- `Porygon z` vs `Porygon-Z`
- `Hoopa hoopa` vs `Hoopa`
- `Type: null` vs `Type: Null`
- `Tapu koko` vs `Tapu Koko`
- `Tapu lele` vs `Tapu Lele`
- `Tapu bulu` vs `Tapu Bulu`
- `Tapu fini` vs `Tapu Fini`
- everything beyond `Melmetal` is missing

Does it even work to have spaces in the names?

I'd like to add the missing stuff did you generated this data from the JSON? Is there some automatic generator or was it done all by hand?

Cheers,
Chris.

ccev:
Apparently Katul lost the script he used to generate monsters.json but I recently made a new one that correctly updates the file. It can be found [here](https://discord.com/channels/358180805031493634/786241014037741609/799061211194589255). Idk when it's going to be used though

philcerf:
@ccev Nice... especially that you generate the data directly from the game master.

Can one also generate the other JSONs from that (i.e. the translations)? I Haven't found a real API description from PogoData.


I generated a current version and and diffed it with the one in the repo,...

Much stuff is **gone**, though,... but this seems to be mostly shadow/purified forms - where I guess they simply don't exist in the master, yet, or were cleaned up again from it?

However, there are some cases:

1. The Normal form seems to be sometimes gone, too, e.g. Chansey.
[...]

PR #278, opened 2021-01-20
Update monsters.json + An update script

ccev:
I recently made a quick Python script to generate the monsters.json based on current GameMaster/Proto/Locale data and then had a discussion about it in #265
This PR replaces monsters.json with the automatically generated file and also puts the script in src/util/, so it's easily accessible.

It'd be pretty easy to also do the locale/ files, though they currently look pretty messy. What's actually being used there?
If there's other files that'd need updating, I could make a script for those as well.


philcerf:
@ccev @jfberry 
I've just generated a new monsters.json with the tool, and then completely went through the diff with the current version in git.

Here's a list of changes:
- most changes are that the new version lacks loads of Normal/Shadow/Purified versions, which are, AFAICS, however all such, that haven't been in the game so far... so that should be fine

Below is the old and new JSON, after merging left and right these above drops of Normal/Shadow/Purified (plus some further tiny changes where the new version got things right, e.g. wrong types in some Alolan forms of the old version.
[old.json.txt](https://github.com/KartulUdus/PoracleJS/files/5846867/old.json.txt)
[new.json.txt](https://github.com/KartulUdus/PoracleJS/files/5846868/new.json.txt)


**You can use it to have a look at the remaining open differences, which are mostly as follows (that is improvements of the new over the old):**


**The boring ones:**
- some pokemon were missing (e.g. Galarian Yamask)
- some types were missing from Alolan/Galarian forms
- some form oderings were from e.g. 555_2343 is now correctly Ice Fire, not the Fire Ice
- some forms were missing (e.g. Delibrid's Winter form or 302_2668 or 613_2672)
- names were corrected (e.g. "Mr mime" -> "Mr. Mime", Ho Oh" -> "Ho-Oh" or "Nidoran male" -> "Nidoran\u2642")
- forms were properly escaped (e.g. "Fall 2019" -> "Fall_2019", or "Exclamation point" -> "Exclamation_point")
- the old one had some weird Mewtwo forms as 151 (Mew) in addition to forms as 150


**More interesting changes:**
- for some Pokemon the base stats changed (e.g. 386, 413, 479 and more)
- in the old JSON, Giratina had a Normal form ("487_1868"), along with Shadow/Purified... but 1868 is neither the ID for Alternate nor for Origin in the new one..(nor in the old one)... no idea where this Normal form came from
- 618_2345 has an emoji change from "\ud83d\uddff" -> "\u26f0\ufe0f"
- Bulbapedia doesn't list Pyroar to have forms, but in the new JSON 668_2587 "Normal" 668_2588 "<no name>" were added.
- Similarly Meowstic has added 678_2589 "Normal" and 678_2590 "<no name>"
- In the old there was 122_31 (Galarian Mr. mime) which now become 122_2339. Similarly there was 77_31 (Galarian Ponyta) which became 77_2336 and 78_31 (Galarian Rapidash) which became 78_2337. From the size of these number I'd guess the new JSON is always right.
- 79_31 (Galarian Slowpoke), where the form ID seems wrong anyway got lost (but it's not in the game yet, so probably fine.
- form IDs were corrected in the new one, e.g. for 25_2669, 25_2670 
- the old one had Shadow/Purified Paras (haven't we really had that already?)... gone in the new none
- the old one had probably erroneous forms 46_996 "Parasect normal", 46_997 "Parasect shadow" and 46_998 "Parasect purified",... however 46 is Paras... those are completely gone from the new one. The same with 599_2195 "Klinklang normal", 599_2196 "Klinklang shadow", 599_2197 "Klinklang purified"... where 599 is Klink, not Klinklang.
- the missing Obstagoon, Perrserker, Sirfetch'd, Mr. Rime and Runerigus were all correclty added... but all with their Normal/Shadow/Purified version in addition, which I think aren't in the game yet (but again, maybe they're just in the master)
 

**Errors (more or less) in the *new* one:**
- The form names "Nidoran_normal", "Nidoran_shadow" and "Nidoran_purified" (for both, male and female)... should probably be Normal/Shadow/Purified.
- Frillish previously had forms: 592_2330 "Female", 593_2331 "Female"  (yes both female)... the new JSON still has the forms with IDs, but the form name is empty (and it also has Shadow/Purified)... that could be an error?
- Some pokemon have their Shadow/Purified forms in the JSON, even though (IIRC) those haven't been in the game (but maybe they just are in the master)... on example is Yamask... just mention it cause maybe it's a bug since Yamask does have Normal and Galarian forms. Wobuffet, Wurmple, Frillish could be other examples of this.
  OTOH, Cubchoo who gained a Winter form in the new JSON, has correctly lost the Purified/Shadow versions in the new one. Same for 94_0, which has added the costume form and has the Normal form, but correctly lost Shadow/Purified (or was there shadow Gengar?)
- Wobuffet lost 202_2328 ("Spring 2020"), Raticate lost 20_2329 ("Spring 2020"), Wurmple lost 265_2327 ("Spring 2020")
- for some the stats got lost, e.g. 679_0, 680_0, 681_0, 710_0, 711_0, 717_0
- Yveltal (717_0), seems to have gotten wrong types (should be Dark Flying according to bulbapedia, which was also the old value)
- some pokemon got lost completely: 718_0 to 807_0
- some forms have empty names, like e.g. 668_2588 or 678_2590


**Kinda not really relevant, AFAIU:**
- For the wildcard-forms like e.g. "351_0" (Castform),... when we actually have subforms (like here sunny, rainy, normal, snowy),... the types make no sense. Like now, Castforms's _0 type was changed from Fire to Normal.. both are kinda wrong.
- The same would be true for everything wit mega evolutions, just that here, the base stats are also changed, aren't they?
- But again, only on forms where there really have multiple forms in the game, cause otherwise _0 is used as Normal, and for that the type/stats is obviously correct.
- another example of the above would be Rotom (type from Water -> Ghost in the _0 form), Darmanitan, Arceus, Shaymin or Meloetta... but in all these cases, the new type (if there "has" to be any)... makes IMO more sense


**Possible improvements:**
- Mewtwo's Armored form is called "A" .. maybe it should get a manual handling of this.
- perhaps escaping Nidoran’s ♀️♂️ with something that's easier writeable by users (e.g. "_f" "_m")?
- Sirfetch\u2019d uses ’ instead of ' (\u0027)... guess no user will recognise this and entered commands would fail... there should be a general rule replaceing \u2019 with \u0027 in all names and form names



So overall it looks pretty good... much better than what we had as the old JSON.
@ccev you might want to check the above notes whether any of these are real issues and whether any of the improvement ideas makes sense to you.

I'll put that now live in our poracle instance and see what happens ;-)

jfberry:
Thanks for diving into this @philcerf - I've never been into the stats side of the game so it helps to have someone with an eye for detail across this!

A few comments:

- Seperation of util.json: This is actually how it was in v3, not sure why Kartul combined it beyond convenience. We absolutely could break it out again but it probably depends on the frequency portions are updated. It's not that big a deal to produce parts and load them in to update if it is not happening very often.
- Some of the slight name changes in the Poracle file will doubtless be to help users actually track for them - "Mr mime" easier to track for (as mr_mime) vs the more proper Mr. mime.  The same would be true of nidoran male being possible to enter vs the more correct unicode version.  
- Forms I imagine equally if using the form tracker.  So our requirements may not be 100% pure in some cases.
- Forms missing - since we're about tracking new things historic missing forms won't matter to us. Odd though.
- The file was previously generated from the game masters (but source lost) -- so the differences in future unpublished items are just changes that we don't care about for sure

Perhaps the tool should be used for a first draft? Or we start to hard code some of the special exceptions

philcerf:
You're welcome :-)

- In any case, I'd suggest to separate generated content from non-generated content, e.g.like Poracle's own translations from the Pokémon name translations.
- For such cases as "Mr. Mime" or also `’` vs `'` ... would it work to somehow define aliases... so that users can do either of "Mr_Mime" or "Mr._Mime"... or "Sirfetch’d", "Sirfetch'd" and "Sirfetchd"? Do you know whether the command parsers would even support "Mr. Mime" (with space)?
For such things one could probably automate the generation of different name variations again, e.g. all combinations of ` ` replaced with `_` + `‘` replaced with `'` ... and `.` replaced with nothing. One Pokemon also has` :` in it's name, I think.
Or maybe it would would be easier, to change poracle to do a fuzzy search.... and give the user a warning like `"Mr_Mime is not a canonical Pokémon name, used "Mr. Mime" instead"`.
The same then for form names.
- Missing historic forms,... probably not the most important thing, yes... but they may return. And I think we even had cases where normal and costume forms where in the game concurrently. So in theory it does in fact make sense to really search specifically for a costume or non-costume form.

I think it's anyway a bit difficult for end users to set up proper rules when there are multiple forms in the game (like Normal + Alolan) plus costumes. Imagine a user just wants to track costumed normal Rattata. _0 would also give Alolan Rattata, so the user needs to specifically add a rule for the very one costume. Same vice versa if he'd just want to track costumed Alolan Rattata. But these cases are probably not that important.

jfberry:
The command line parser has a underscore to space converter - so you can do rare_candy or mr_mime but I agree aliases would make sense here.  The current multi language patch I've done translates everything from multiple forms back to a single one for command parameters and I can see how it could be expanded to manage aliases also... (ponders)

philcerf:
But that converter goes only in one direction I assume, right?
I.e. one couldn't do `/track Mr. Mime d100`?

jfberry:
It displays as it is in the file, you track with an underscore instead of a space

philcerf:
Hey guys :-)

I just did some further checking (with the generated JSON).
Something like:
`/track Icognito formExclamation_point`
will fail.
Seems to affect everything with an underscore.

Any ideas?

Cheers.

philcerf:
Also, any thought already, on how monster name translation should happen? (i.e. like in a separate file or in `config/locale/<lang>.json`?

Cause, unless @ccev is already on it, I'd might at least try to produce something for that.

philcerf:
If one changes the form name within the generated JSON to use a space e.g. `Question mark` instead of `Question_mark` the command:
`/track Icognito formQuestion_mark` succeeds (while /track `Icognito formQuestion mark` obviously does not.

Guess the problem is that poracle itself mangles up the names? Or at least I'd say it's better if the escaping of space in form names would be handled (once) in the generator-tool... and not in poracle itself.

Petap0w:
> If one changes the form name within the generated JSON to use a space e.g. `Question mark` instead of `Question_mark` the command:
> `/track Icognito formQuestion_mark` succeeds (while /track `Icognito formQuestion mark` obviously does not.
> 
> Guess the problem is that poracle itself mangles up the names? Or at least I'd say it's better if the escaping of space in form names would be handled (once) in the generator-tool... and not in poracle itself.

Poracle has to be given spaces replaced with underscore or there is no way for it to know when to link two words separated by a space in a command (part of same "double word"? two parameters?).

philcerf:
> Poracle has to be given spaces replaced with underscore or there is no way for it to know when to link two words separated by a space in a command (part of same "double word"? two parameters?).

it's clear that the comment mustn't have space.  But I'm talking about, that **the form name in the JSON** doesn't work if it has underscores (i.e. not spaces).
And that's kinda strange, IMO.

Petap0w:
> > Poracle has to be given spaces replaced with underscore or there is no way for it to know when to link two words separated by a space in a command (part of same "double word"? two parameters?).
> 
> it's clear that the command mustn't have spaces. But I'm talking about, that **the form name in the JSON** doesn't work if it **has underscores (i.e. not spaces)**.
> And that's kinda strange, IMO.

yes then the tool that generates content for Poracle has to generate compatible content, ie. with underscores replaced by spaces and that'll work well then.

Petap0w:
Who would be the lead on this PR so we can see how to move forward and get this implemented soon? ccev? philcerf?
That way I could review the rest of the code to match new source files available.

ccev:
Currently occupied with other stuff, I'll leave everything up to you guys. Just here to help with the update script. If you need help with pogodata, feel free to DM me on Discord.

philcerf:
> yes then the tool that generates content for Poracle has to generate compatible content, ie. with underscores replaced by spaces and that'll work well then.

Well my blind assumption would be, that there is some code in Poracle itself, which mangles up the names, i.e. changes the version **with** underscores (and vice versa), that it gets from the generated files (e.g. `Exclamation_Point`) to one with spaces, which subsequentially fails to work.

Wouldn't it be a much better approach to simply stop Poracle from changing the names and taking them as is, if the newly generated JSONs have anyway the correct form and don't use spaces anymore?

Did some blind grepping on the sources and it seems one or both of these could be the "offenders":
```
PoracleJS/src$ grep -n -R "\.replace" | grep _
init/migrateV3.js:61:				obj[key.replace(/([A-Z])/g, '_$1').toLowerCase()] = obj[key]
lib/discord/commando/commands/area.js:7:		const confAreas = client.geofence.map((area) => area.name.toLowerCase().replace(/ /gi, '_')).sort()
lib/discord/commando/commands/area.js:8:		const confAreas2 = client.geofence.map((area) => area.name.replace(/ /gi, '_')).sort()
lib/discord/commando/commands/area.js:38:		const areaArgs = args.map((a) => a.replace(/ /g, '_'))
lib/discord/commando/commands/area.js:42:				const oldArea = JSON.parse(human.area.split()).map((area) => area.replace(/ /gi, '_'))
lib/discord/commando/commands/area.js:62:				const oldArea = JSON.parse(human.area.split()).map((area) => area.replace(/ /gi, '_'))
lib/discord/commando/events/message.js:9:	args = args.map((arg) => client.translator.reverse(arg.toLowerCase().replace(/_/g, ' '), true).toLowerCase())
lib/telegram/middleware/commandParser.js:15:			return args.map((argss) => argss.map((arg) => translator.translate(arg.toLowerCase().replace('_', ' '))).filter((x) => x))
```

In particular, the later two ones (`replace(/_/g, ' ')`) which actually go from `_` to ` `.

Cheers.

jfberry:
I think you might be misunderstanding what's going on.
The Pokemon names/forms in the monster file are 'normal' because that's needed for translation and on-screen display.
Poracle splits commands on spaces into an array.
Then it replaces the _ with spaces so that the pokemon name (or form or quest type) with an underscore now is a single string with a space. This then allows reverse translation back to english and matching pokemon to match the display names.

Ideally I'd like poracle to support quotes as well - eg /quest "rare candy" as well as /quest rare_candy and have taken the first step to changing the command line processor in the multilang pr

philcerf:
Hm, I see. Well if it's used for displaying, that obviously makes perfectly sense then.

But I guess Poracle should then somehow also handle cases that users can use `Sirfetch'd` (U+0027) or even `Sirfetchd` intead of `Sirfetch’d` (U+2019), which will be in the JSON, right?
Cause a normal user likely never finds `’` and just removing them will probably also confuse people.
Any ideas whether/how you'd like to handle this?


But from the script's PoV, it would needed to be changed to replace any `_` in form names (only) to ` `, right?
Or @ccev, do you know whether there's a field retrievable with PogoData, data shows the "pretty" names?

It seems one cannot get it from the mon object itself, there's only m.template, which you already use.
https://github.com/PokeMiners/pogo_assets/blob/master/Texts/Latest%20APK/JSON/i18n_english.json contains e.g.
`"form_rain", "Rainy",`
which is, I guess, what we'd really want.. but I found no way to map from the Pokemon to `form_rain`

Cheers,
Chris

jfberry:
Perhaps we should move to have our own alias list defined outside of this? You can also track on pokedex no of course (and through poracleweb)...

philcerf:
hmm seems pretty much of an effort especially when you have to support this list for n langauges.

philcerf:
And I'd expect it's effectively not that much mapping which needs to be done (from the entered command to the actual value looked for in monsters.json and friends):
- `_` to space (which would however get a problem, should Niantic ever add a Pokemon or form name that really contains a `_`)
- counterparts like `'` which users can enter more easily to their unicode chars like `’`

The whole schema, if automatic handling and not some static list, was used, would however break once we have a replacement char in our commands (like the `_`) for which we'd have more than one possible mapping on the backend side.

jfberry:
Thanks Malte and phil for this.  As part of the multilang PR we (well petap0w really!) has taken this concept and run with it to generate the monsters, moves and associated translation files.  This PR will contain the generator and outputs so we'll close this PR. @philcerf would appreciate if you are able to provide some validation that the new files we have generated are right or not so that we can include the necessary overrides

philcerf:
Will this already include that joint effort to unify all that stuff over different projects, which you've mentioned on discord?

And is the new code already available?

Petap0w:
> Will this already include that joint effort to unify all that stuff over different projects, which you've mentioned on discord?
> 
> And is the new code already available?

check out current PR for multilang, it's in there

jfberry:
> Will this already include that joint effort to unify all that stuff over different projects, which you've mentioned on discord?
This is primarily around icon repository format currently, and I don't think it has settled yet

philcerf:
Does that require Node 12 now? Not yet in Debian :-(

And how would I run the tool to generated a new monsters.json?

Petap0w:
Run npm install for new dependencies then go in src/util and node updateGameData.js

philcerf:
@Petap0w I manually went through all changes between my last version generated with Malte's tool (which in turn I manually diffed with the one from the git - see above).

These are the notable changes:
rather boring:
- color/emjoi lines gone in new
- from "Exclamation point" to "Exclamation Point" ... same for Unkown-?
- Swinub/Piloswine/Nosepass/Aron/Lairon/Aggron/Mamoswine/Probopass Normal/Purified/Shadow/Lileep/Cradily/Anorith/Armaldo/Spheal/Sealeo/Walrein added (I didn't recreate my old monsters.jsons from @ccev's tool)
- "Adventure_hat_2020" -> "Adventure Hat 2020"
- "Nidoran normal" -> "Normal", for both, male/female and all forms (i.e. the same for shadow/purified)
- "West sea" ->  "West Sea" (same for east)
- "Red striped" -> "Red Striped" (same for blue)
- "Galarian standard" -> "Galarian Standard" (same for Zen)

perhaps interesting:
- "proto": "EXEGGCUTE_SHADOW" lines added in new, some are empty though
- in old, MANY(!!!), but not all, _0 entries had form/name : ""... which was changed in many cases to "Normal", not sure if this makes sense, since _0 is the wildcard for all forms?
  For example, "100_0" had it changed, while for "102_0" it stayed empty
  I could imagine (but took only some random samples) that any n_ with no real further forms, has it set to "Normal" and all those which do have actually more forms, has it left ""?
  also, the actual normal n_m forms got "default_form": true, added
- Zygarde was added (?)... but with empty stats
- for Yveltal the type is now right, the base stats are not longer empty (didn't check them though)

errors / things to look at:
- "Sirfetch\u2019d" -> "Sirfetchd"
  => as I've said previously, I think we should allow all possible aliases e.g. "Sirfetchd" "Sirfetch_d" "Sirfetch'd" and "Sirfetch’d"
- for the following forms, the form name changed from "Normal" to "Female"
  "592_2330"  593_2331  668_2588 678_2590


So other then the issues mentioned right above it looks quite good.

But obviously, all possible issues I've brought up in https://github.com/KartulUdus/PoracleJS/pull/278#issuecomment-764279735 above, may still apply to the generated monsters.jsons from your code.
So you may want to check these as well.

philcerf:
I also did a very basic check of the generated locale/pokemonNames_de.json, but only to the extent of checking that the errors I'd had previously corrected in PR #252 are still correct.

Petap0w:
> "Nidoran normal" -> "Normal", for both, male/female and all forms (i.e. the same for shadow/purified)

don't get that one? normal nido have "Normal" as form name, and shadow/purified have Shadow or Purified, all correct?

It's normal that all pokémon without any form out for now have _0 labelled as Normal as that's the form they have. When they have more than that each one has its tag, the default in proto is indicated, and _0 becomes a fallback.
> Sirfetchd

You can always add all the cases you wish in your custom.locale.json if it's too complicated for the users to use simply Sirfetchd - all "translations" will be working for command parsing, even "Sirfetchd":"thatpokémonwiththewhateverinthename"
>"592_2330" 593_2331 668_2588 678_2590

the name didn't change from normal to female: they have a Normal form (it's there 592_2168 593_2171), and a form that is "Female" form so it's there as well with the name from the game.

philcerf:
> don't get that one? normal nido have "Normal" as form name, and shadow/purified have Shadow or Purified, all correct?

Yes, I really just listed **all** changes, even the correcting ones, so that we can see what got better from the previous generated files.

> It's normal that all pokémon without any form out for now have _0 labelled as Normal as that's the form they have. When they have more than that each one has its tag, the default in proto is indicated, and _0 becomes a fallback.

Hmm the only problem I can see here is kinda the following:
What if a user, sets a rule to explicitly match formNormal, before a further forms exist?
It would get stored as the wildcard rule, I assume... i.e. the user would especially get no error (that no such form exists) or warning that it's used for the wildcard form... should other forms be added later on, his (explicit normal-only) rule would then suddenly match all forms.
Still much better than the other way round (i.e. a rule that specified no form, should never got mapped to the actual normal form version (if there is only the normal form) but only to the wildcard version)... but might still cause confusion.

> You can always add all the cases you wish in your custom.locale.json if it's too complicated for the users to use simply Sirfetchd - all "translations" will be working for command parsing, even "Sirfetchd":"thatpokémonwiththewhateverinthename"

Ah I see, that's nice indeed... can / shouldn't we provide that out-of-the-box for these typical cases like Sirfetched' or Type: Null ?

I could help to write an according patch.
You mean PoracleJS/config/locale/*.json I assume?
How would that work, if one want's to give multiple alternatives?
I mean I could set only one:
`        "Sirfetchd":"Sirfetch_d",`
and not more?

>  "592_2330" 593_2331 668_2588 678_2590

The only strange thing I've noted with these was, that Bulbapedia doesn't seem to list them as having forms:
https://bulbapedia.bulbagarden.net/wiki/Frillish_(Pok%C3%A9mon)
https://bulbapedia.bulbagarden.net/wiki/Jellicent_(Pok%C3%A9mon)
https://bulbapedia.bulbagarden.net/wiki/Meowstic_(Pok%C3%A9mon)
https://bulbapedia.bulbagarden.net/wiki/Pyroar_(Pok%C3%A9mon)
https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_with_form_differences

It seems they count it just as plain gender difference. If PoGo handles that different, well we're probably bound to keep it. Or alternatively, if Normal-form is always male and Female-form is always female, we could perhaps add special handling that would allow users to do not just:
`/track Frillish formNormal`
`/track Frillish formFemale`
but also the more usual:
`/track Frillish male`
`/track Frillish femal`
and internally map the later to the former.


Cheers

Petap0w:
I mean `config>custom.fr.json` that you can use to add your custom translation on top of the regular` config>locale>fr.json` (new addition with multilang processing).
There you can put `"Sirfetchd":"Sirfetch d"` (no _ since those are already converted to space when command is parsed) then a user making `!track sirfetch_d` would register correct tracking for sirfetchd.

As for the rest, the file is really game master file (ie. direct game data files) parsed and converted to Poracle format - so like you say not much latitude to try change things since in the end the game dictates what should be there.

I think we have approached this from every angle and it's now quite good to go

philcerf:
I see.

One last thing, with current develop (99ec950a16bc1d3c926ebe3be1c0bceb7b2a8821) the following don't work:
```
 [02:49]  PoGo MUC Poracle <<< /track Schalellos formWest_Sea iv100 d2500
 [02:49]  PoGo MUC Poracle >>> 404 Keine Monster gefunden
 [02:49]  PoGo MUC Poracle <<< /track Gastrodon formWest_Sea iv100 d2500
 [02:49]  PoGo MUC Poracle >>> 404 Keine Monster gefunden
 [02:50]  PoGo MUC Poracle <<< /track Barschuft formBlue_Striped
 [02:50]  PoGo MUC Poracle >>> 404 Keine Monster gefunden
 [02:50]  PoGo MUC Poracle <<< /track Flampivian formGalarian_Standard iv100
 [02:50]  PoGo MUC Poracle >>> 404 Keine Monster gefunden
 [02:50]  PoGo MUC Poracle <<< /track Flampivian formGalarian_Zen iv100
 [02:50]  PoGo MUC Poracle >>> 404 Keine Monster gefunden
```
(same with English monster names)

AFAICS, all the forms are correctly in monsters.json,... and the monster names themselves are also there, e.g.
```
 [02:50]  PoGo MUC Poracle <<< /track Flampivian formStandard iv100
 [02:50]  PoGo MUC Poracle >>> 👌
 [02:50]  PoGo MUC Poracle <<< /track Flampivian formZen iv100
 [02:50]  PoGo MUC Poracle >>> 👌
```
work.

jfberry:
Have pushed a fix for this to develop
By side effect it may also fix template names with underscores (use a space in the dts file) - I haven't tried this but think the same fix would work there

philcerf:
Thanks, it fixed the issue with the form names, an I think also with the template names (they're now set, with " " in the DB, where one had _ in the command.
Just didn't rename my templates again to verify, whether they really match.

PR #293, opened 2021-01-26
Bypass default distance settings for admin

Petap0w:


jfberry:
Was included in fixadmin branch

PR #298, opened 2021-01-27
Choose multistaticmap, staticmap or none for notifications

LordJavi:
Use staticmap, multistaticmap or "" for disable it

⚠️ Requires SwiftTileserverCache >= 2.0.0 ⚠️

Thanks @ccev for multistatic map template

68747470733a2f2f6d656469612e646973636f72646170702e6e65742f6174746163686d656e74732f3637363832353133383736323837343931302f3734393232393237323935383336393836322f6d756c7469746573742e706e67

 https://gist.github.com/ccev/47b6de2a2f4578a06d14058f323ba0ba


jfberry:
What a great PR, thanks!
We've made quite a few changes around these same files in the multilang PR so this will need to queue up for final merge after that you can rebase - can you post some example outputs to the discord so we can all see?

config/default.json
```diff
@@ -112,6 +112,14 @@
         "staticProviderURL": "Your TileServerCache Server Address",
         "geocodingKey":["Your Google Geocoding Key if you Use google as provider"],
         "staticKey":["Your MapQuest or Google Key"],
+        "multistaticmaps": {
```
jfberry, line 115: There is also a request to disable static map generation for people who don't want to use them for certain types.
It would be cool if this were:
`
staticmaps: {
  "pokemon": "multi",
  "raid": "standard",
  "invasion": "none"
}
`
or something similar

LordJavi: copy that! I'll update branch asap 😃

jfberry:
I am planning to merge the big multilang on sunday, then will be able to get this in as next job.  It will need rebasing I think and a couple of changes before I can merge it.
1. Please make the default.json defaults so everything is as normal (ie using regular static maps)
2. Please keep the original template names for regular static maps so users don't need to rename their custom templates

This is so everything works for people that just pull to update, just makes my life easier responding to people on discord!

Oh, and I think this is excellent by the way I will be using it! :-)

jfberry:
I am through my backlog of changes, so I will happily look at this and fix it up as I suggested next week if you don't have time.

jfberry:
Merged into performance PR, thanks :-)

PR #303, opened 2021-01-28
Add Caches and logs as volumes

muckelba:
so that they dont get deleted when the container stops


jfberry:
the new clean cache is called .cleancache* (one per connection), can wildcards work or should I put them in a folder? (could decide now and get this in this pr ready!)

muckelba:
wildcards will not work. i'd suggest to merge this now and adjust this file when PRing the move to a cache directory

jfberry:
Ok I will make sure that I remember to move to a cache directory (I will change my code now so I don't forget)

PR #306, opened 2021-01-29
Adjust User PvP Ranking DTS

plinytheelder:
adjust user PvP ranking DTS since only one value is returned when notifying users, which is troublesome if a mon is say a Rank 1 for both Great and Ultra League.

Still allows for single template use for a monster IV template, and can be used to only display PvP stats if the user actually cares and subscribed to them.

ex.

{{#compare pvpUserRanking '>' 0}}
add your fancy pvp DTS here...
{{/compare}}


jfberry:
The problem with this is the way that the alerts translate through to the production of the templates.
There is duplicate elimination in two places - first at the database level, and then at the code level meaning that if the user matches on two conditions at once you can't guarantee which will be passed through to the template production.

For this to actually work reliably it probably requires a change in logic:
eg: 
a) remove database level duplicate elimination (ie the group by)
b) change the code level duplicate elimination to promote some records over others in a rules based way - eg if there are data items like this that need to be promoted to template processing that exist in one but not the other, accept that record

Incidentally, a similar problem exists if a match occurs with two different templates - it would be random which one you got. So it would be interesting to see if the result should be a 'merged' record or a single record wins (like now)

For this reason a global restriction to the template list probably makes more sense (which was the pvpFilter PR that separately looked at restricting this list)

philcerf:
@jfberry and do you see any easy way to make it working? Cause IMO that would be really interesting, as right now (see https://discord.com/channels/358180805031493634/533927791096233984/809563811522674699) the whole mechanism is kinda limited.

jfberry:
This is no worse than previous so if rebased I will accept it.
Issue is that it doesn't really solve the problem in a multiple alert situation, I would prefer to see a solution which though about what to do when multiple alerts are fired and this is squashed (eg bringing some user parameters forward in preference to others). Templates is another example which wouldn't work properly in this situation but in both of these I guess you could say 'whatever'

philcerf:
@plinytheelder If you rebase, could you then also add the same code for the user selected greatcp / ultracp?

@jfberry What exactly do you mean by multi alert situation? I thought it would be about like that: a Pokemon appears, it goes through the list of filters for that user and the first one that matches selects things like the template and so on.

So if e.g. a user selects:
/track everything great100 templateFOO
/track Snorlax great1 templateBAR

And the order is preserved  and a Snorlax with rank 1 appears,... one would get templateFOO?

One might perhaps improve on that, if the matcher itself would go on, and select the most specific match (but that would also put more load on the DB)... and opens the question on what actually is considered as "the most specific".

But for PvP, and specifically for the problem here, I don't think it's so relevant (at least if I understand the problem correctly), cause regardless of by which of the above two rules the Snorlax would be selected, the user likely would want PvP data to show up in the output.

jfberry:
The database does not guarantee any order
Poracle squashes duplicates, so you will only get one alert.
Which template you get in that situation will be random
...and the way this patch is written if one was not a great league match you wouldn't end up with your custom match details being available in DTS (eg if one was just !track snorlax with no pvp)

philcerf:
ah... now I understand...
so basically, if one had:
/track everything iv100
/track Bronzor great1 greatcp600

where Bronzor itself would be at rank1 (and IV100%) with 689,... one could end up not getting the PvP ouput, when it was selected via the wildcard

philcerf:
That's indeed unfortunate....do you see any solution?

I mean for the PvP case, one could probably, after a match **without PvP values set** go on and look for any further records which would match and which **have PvP values set**, and if so, send these along.
Would be not were performant but at least work...

But for the same problem with templates there is no real solution, cause there is no answer which should be preferred.

jfberry:
Well all the results come back in this case, in code you could have some rules about what gets promoted during de-duplication. Either preferring some kinds of match, or just propagating some user fields.
Templates probably doesn't matter that much as you say - it's just an example of how this was always there but until this and the predecessor patch that used user variables from the match - it wasn't something that would drastically affect output (since both would have been right before).
But filtering to 0 on a PVP mon because of a second non-pvp match would not be good
And I can also see why for the majority of cases this is not an issue... which is why I said I will accept the PR but this is not a complete solution

philcerf:
Well I'd love to see this feature implemented, but I think that's a bit too deep in the core code for me to handle this right now (without breaking everything else ^^)... do you see any chance to put it on your bucket list?

With templates it's as least so far a problem, that if the wrong template was selected, it may not even contain rules for printing the PvP stuff... which actually sounds reasonable, e.g. I have made templates for PvP + no-PvP... OTOH though... I only made this because I "suffered" from the problem which would be fixed by the ideas here.

So once it would be guaranteed, that the right user rank/CP values go into the DTS... there's no strong need anymore, to create a DTS without PvP printing.

The only solution I'd see for the template selection, was to set some kind of priority on the templates, i.e. if multiple apply the one with the highest prio wins.


> And I can also see why for the majority of cases this is not an issue... which is why I said I will accept the PR but this is not a complete solution

hmm I'm not that sure anymore myself... I mean there's a good chance that this affects at least all those cases, where the high ranking PvP values are close to IV100%, cause that's what people likely filter for, too, without PvP in mind.

plinytheelder (commented):
rebased...I think

PR #312, opened 2021-02-03
Update grunts

Petap0w:


calestyo:
Would be nice if we could get that also auto-generated from the master. Have you seen Malte's tool for RocketMAD? (https://github.com/cecpk/RocketMAD/pull/198)

jfberry:
Leaving this PR here for a bit if people pull it, it is for pre-multilang.
multilang has new grunts also in a seperate file
I believe petap0w does generate this from malte's file (which does not come from gamemaster but is manually curated)

jfberry:
Multilang seems stable enough, people can come here to get updated grunts - thanks as usual!

PR #316, opened 2021-02-07
Remember messages to be deleted and event cache between restarts

jfberry:
Details of messages to be deleted and current webhook events will be stored in the .cache directory on shutdown. This will be reloaded on startup.


hamster007Github:
Cleaning and avoiding duplicates after restart working fine on first test with Telegram 👌

hamster007Github:
Running now for 2 days with telegram without any issues. No duplicates and cleaning messages correctly after restart.

jfberry:
How many times have you restarted? Do you have any discord clean commands set up?

hamster007Github:
Have groups and private chats with active cleaning (pokemon). Restarted 5...6 times with active tracked pokemon. But Telegram only.

Issue #320, opened 2021-02-09
need contextual help with translations


calestyo:
Hey.

Based on `PoracleJS/config/locale/fr.json` I've started with the German translation, which, as of now is in the following state:

My problem is, that for many strings the context is pretty much unclear (even when looking for it in the sources).
Especially:

1) Are the following all (only) parts of commands respectively their parameters (i.e. **not** a command’s output)?
- `add`
- `remove`
- `level`
- `template`
- `d`
- `female`
- `male`
- `genderless`
- `gen`
- `everything`
- `individually`
- `maxlevel`
- `great`
- `greatcp`
- `ultra`
- `ultracp`
- `maxcp`
- `maxiv`
- `maxweight`
- `maxatk`
- `maxdef`
- `maxsta`
- `cp`
- `iv`
- `atk`
- `def`
- `sta`
- `weight`
- `language`
- `area`
- `egg`
- `help`
- `invasion`
- `location`
- `poracle`
- `quest`
- `raid`
- `start`
- `stop`
- `track`
- `tracked`
- `unregister`
- `untrack`
- `weather`
- `list`
- `instinct`
- `valor`
- `mystic`

At least this:
- `form`
seems to be used in both command and text-output context, in which case I would write once "form" and once "Form" in German.

2. I assume these is all just the weather forms:
- `sunny`
- `rain`
- `partly cloudy`
- `cloudy`
- `windy`
- `snow`
- `fog`

But isn't there a different name for `sunny`, depending on whether it's day or night?

3. I assume these are all used for quests:
 - `Poke Ball`
- `Great Ball`
- `Ultra Ball`
- `Potion`
- `Super Potion`
- `Hyper Potion`
- `Max Potion`
- `Revive`
- `Max Revive`
- `Incense Ordinary`
- `Normal Lure Module`
- `Glacial Lure Module`
- `Mossy Lure Module`
- `Magnetic Lure Module`
- `Razz Berry`
- `Nanab Berry`
- `Pinap Berry`
- `Golden Razz Berry`
- `Silver Pinap Berry`
- `Poffin`
- `Sun Stone`
- `Kings Rock`
- `Metal Coat`
- `Dragon Scale`
- `Up Grade`
- `Gen4 evolution stone`
- `Gen5 evolution stone`
- `Fast TM`
- `Charged TM`
- `Rare Candy`

But are they used solely for output or also for any commands?

4. For they invasion types, there seem to be no translations for:
[...]

PR #323, opened 2021-02-09
Localisation de for default dts

hamster007Github:
Adding/extend german localisation

## Description
Added german templates to default dts.json
Extend german localisation file de.json based on fr.json (without quest parts)
Remark: Someone with Discord should test "de" DTS templates for discord.

## Motivation and Context
German support missing for DTS and many phrases in localisation file

## How Has This Been Tested?
local setup with pokemon tracking+ weather change on Telegram. Rest untested.

## Screenshots (if appropriate):
-

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #324, opened 2021-02-09
updated German locale

philcerf:
This was made besed upon config/locale/fr.json, however, with all quest texts
removed.

As far as applicable, I've used the “offical“ spellings as found in the game.
This goes so far, that I've translated “genderless” with „geschlechtunbekannt“,
which no-one would use in German, however, „geschlechtunbekannt“ is used by the
game as search-keyword, so I stayed with that.

Also, as described in https://github.com/KartulUdus/PoracleJS/issues/320 I lack
context for a number of strings.

Commands should IMO be written lower-case, even in German.
As far as I could identify strings belonging to commands, I left them
lower-case. Future authors are invited to improve upon that.

Further, I've tried to really translate most things, even if it’s strange.
For example, most Germans are probably more used to using English commands like
“/area list” instead of something like “/gebiet auflisten”, but since the
English commands should still work, even when language is set to German, I think
it’s okay to pull it through.

## Types of changes

## Checklist:


jfberry:
See #323 please work out between yourselves which one will be master

philcerf:
Please wait with merging this,... I'll provide an updated PR later.

philcerf:
Based on some ideas in #323 and discussion with @hamster007Github I've revised my locale:
- some minor changes like the (presumably) official name of the Pikachu Adventure Hat 2020 form (which seems to be Explorer Pikachu in English)
- removed some more strings which seem to be no longer used
- For error messages and the like, I've replaced the `404` (which doesn't really make sense here) with the German version of `Error`.
- Further I've moved away from "formal/polite addressing" (of people) which I've had used in my original version. I did however not chose the more personal form that @hamster007Github used in his, either, but completely avoided it at all by not directly addressing people.

@jfberry I think you can merge this now... but actually I came the conclusion that the English texts should perhaps be revised as well, as a number of them is rather vague or sloppy.

config/locale/de.json
```diff
@@ -1,39 +1,204 @@
...
+	"does not seem to be registered. add it with":"scheint nicht registriert zu sein. Hinzufügen mittels",
+	"webhook":"webhook",
+	"add":"hinzufügen",
+	"<Your-Webhook-url>":"<Webhook-URL>",
+	"channel":"kanal",
+	"no valid areas there, please use one of":"Fehler: Kein gültiges Gebiet angegeben. Folgende Gebiete sind verfügbar:",
+	"Added areas:":"Hinzgefügte Gebiete:",
+	"Removed areas:":"Entfernte Gebiete",
+	"Current configured areas are:":"Aktuell sind folgende Gebiete festgelegt:",
+	"nameYourWebhookName":"webhookName",
```
jfberry, line 29: the name in front here is the command parameter so needs to be there
eg namewebhookName

config/locale/de.json
```diff
@@ -1,39 +1,204 @@
...
+	"disabled":"deaktiviert.",
+	"You are currently set to receive alarms in":"Folgende Gebiete sind gesetzt:",
+	"You have not selected any area yet":"Es wurde kein Gebiet gesetzt.",
+	"You're tracking the following monsters:":"Folgende Pokémon werden beobachtet:",
+	"You're not tracking any monsters":"Es werden keine Pokémon beobachtet.",
+	"Everything":"_alle_",
+	"any":"_alle_",
+	"greatpvp":"Superliga",
+	"ultrapvp":"Hyperliga",
+	"gender":"Geschlecht",
```
jfberry, line 95: this is a command parameter, so should be lower case

config/locale/de.json
```diff
@@ -1,39 +1,204 @@
...
+	"Your alerts are currently":"Benachrichtigungen sind aktuell",
+	"enabled":"aktiviert.",
+	"disabled":"deaktiviert.",
+	"You are currently set to receive alarms in":"Folgende Gebiete sind gesetzt:",
+	"You have not selected any area yet":"Es wurde kein Gebiet gesetzt.",
+	"You're tracking the following monsters:":"Folgende Pokémon werden beobachtet:",
+	"You're not tracking any monsters":"Es werden keine Pokémon beobachtet.",
+	"Everything":"_alle_",
+	"any":"_alle_",
+	"greatpvp":"Superliga",
```
jfberry, line 93: this is a command parameter

config/locale/de.json
```diff
@@ -1,39 +1,204 @@
...
+	"raid":"raid",
+	"start":"start",
+	"stop":"stop",
+	"track":"beobachte",
+	"tracked":"status",
+	"unregister":"abmelden",
+	"untrack":"unbeobachte",
+	"weather":"wetter",
+	"list":"auflisten",
+	"instinct":"Intuition",
```
jfberry, line 124: i think this is a parameter for gyms (as with valor and mystic)

jfberry (commented):
Looking good just a few points

config/locale/de.json
philcerf: Sure about that? All 4 seem to be only used in the output of /tracking ...?

config/locale/de.json
philcerf: same here... isn't the parameter greatN and greatcpN?  AFAICS, greatpvp is used in the output string of the /tracking command

config/locale/de.json
philcerf: you're right,... the three team names aren't used in any other place, AFAICS,... so will me lower case in my revised commit

config/locale/de.json
philcerf: mhh not sure wheter I really understand this, but done anyway.

btw: that's one of the occasions where something isn't translated:

```
return await msg.reply(`Webhook ${target.name} ${client.translator.translate('does not seem to be registered. add it with')} ${client.config.discord.prefix}${client.translator.translate('webhook')} ${client.translator.translate('add')} ${client.translator.translate('<Your-Webhook-url>')} ${client.translator.translate('nameYourWebhookName')} nameYourWebhookName`)
```

philcerf:
In some (not yet pushed) commits I've added the translations for all the form names and stumbled over something:

First, there may be multiple `"Normal": "…"` in a locale,... one for the type another one for the form… and I could imagine that these translate to different strings in different languages… wouldn't that cause problems?

Second, take the following (which uses the official German translations):
```
        "Standard":"Normalmodus",
```
If I'd want to abbreviate this (which belongs to Darmantian) to just `Normal`,... I'd probably already into troubles, wouldn't I?
But of course we could say: just don't do it and leave the "modus" (=mode) in.

However, looking at the following:
```
        "Ordinary":"Standardform",
```
It would be pretty ugly to leave this at `Standardform`, because then people would need to use `formStandardform` having the "form" in twice.
But if I reduce it just `Standard` how could Poracle possibly which I mean, the English one as used for Darmantian’s form or the German one?

jfberry:
> In some (not yet pushed) commits I've added the translations for all the form names and stumbled over something:
> 
> First, there may be multiple `"Normal": "…"` in a locale,... one for the type another one for the form… and I could imagine that these translate to different strings in different languages… wouldn't that cause problems?
> 
> Second, take the following (which uses the official German translations):
> 
> ```
>         "Standard":"Normalmodus",
> ```
> 
> If I'd want to abbreviate this (which belongs to Darmantian) to just `Normal`,... I'd probably already into troubles, wouldn't I?
> But of course we could say: just don't do it and leave the "modus" (=mode) in.
> 
> However, looking at the following:
> 
> ```
>         "Ordinary":"Standardform",
> ```
> 
> It would be pretty ugly to leave this at `Standardform`, because then people would need to use `formStandardform` having the "form" in twice.
> But if I reduce it just `Standard` how could Poracle possibly which I mean, the English one as used for Darmantian’s form or the German one?

Poracle back converts everything to english for command processing, so the forms will come back to english. So in the above case Standardform would be seen as Ordinary by the track command parser.
So in the case where there is something like normal as a destination in german for something that isn't normal in English you can imagine anything can happen. And the behaviour might be different for telegram and discord as these are different routines.  I can imagine some prioritisation could be built into this but optionality probably can't without a significant rewrite - best try and avoid

philcerf:
Hmm I guess that could really get us into troubles sooner or later, cause it's difficult for translators to really check that.
And we always have the possibility that TPC/Nia come up with names/translations for which it is basically unavoidable.

How easy/desirable would the following be to add to your multi-lang stuff?
1. allow users to set three languages (maybe with some default that sets all): 
- one for just the commands (i.e. whether it's `/egg` or `/ei`)
- and for other command parameters (like `levelN`, `greatN`, `mystic`, `ex`, etc.)
2. one for Pokemon/form/quest names
3. one for just the output texts (except the names from (2), so e.g. error messages and the like

And then changing the logic, that the command parser really expects these languages i.e. one can no longer use Bulbasaur if the language for (2) was German (where it would be Bisasam).

That would probably solve any such possible ambiguities, cause it's clear, whether the parser must translate or not, and thus it's clear whether e.g. `Standard` would be German and actually means `Ordinary` or whether it was English and means `Standard` (as in Darmantian's form).

Cause we cannot even use the `form`XXX prefix to make it clear what language something is, since e.g. it's the same in EN or DE ... and one would need to make hacks like artificially translating it as DEform or so.

philcerf:
btw: I'm also thinking about cases like a Pokemon with the name `Mystic` would be released... which would make commands ambiguous whether it's the Pokemon or the gym team.

philcerf:
A solution for the later could be to make the command structure more strict, e.g.
/raid <boss-name>... <conditions|flags>...

Where:
- boss-name needs to come first
- and if multiple boss-names where to be given, one would need to make special syntax like boss,boss,boss or [boss boss boss] ro so, so that one can identify where the bosses end and where the conditions/flags start.

PR #326, opened 2021-02-10
Improve everything permissions

philcerf:
As per discussion in discord.

Perhaps you may want to have different option names, e.g. instead of `allow-everything-only` the name `allow-wildcard-only` or something like this could be a reasonable choice?

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -79,8 +79,7 @@
         "pvpFilterUltraMinCP": 2450
     },
     "tracking": {
-        "disableEverythingTracking": false,
-        "forceEverythingSeparately": false,
+        "everythingFlagPermissions": "allow-any",
         "defaultDistance": 0,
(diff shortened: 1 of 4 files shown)


jfberry:
This tidies these options up nicely.  In the new logging PR I have checks for valid options selected and a warn on startup. So I think that this would make sense to include these and would proposing queuing this until post logging changes merged.
This is also a good place to provide a backward compatibility setting - detecting the old settings and setting in memory the right setting and also providing a warning on startup that a deprecated config option has been used

philcerf:
this now also fixes #327

jfberry:
Now logging is patched you can see how to have poracle check for a valid entry in the config file and warn

philcerf:
rebased and added the checking

philcerf:
rebased and new option names

PR #332, opened 2021-02-11
allow to use team colours rather than names in !raid and !egg commands

philcerf:


philcerf:
if you should add translations for the other locales,... make sure to put them in the same place than I did here: https://github.com/KartulUdus/PoracleJS/pull/324/commits/13d5951c27855ccc3af642d801aa49c47c57cd4a

PR #333, opened 2021-02-11
manually set the proper form name for Armored Mewtwo during generation

philcerf:
This is not just needed because `Armored` is the form name users will expect, but also because with just `A`, one couldn’t localise it, since Unown A shares the same form name.

Which also kinda points us to a more general problem in the whole architecture:
What if Niantic adds form names which are identical in English, but translate differently in other languages.

@Petap0w would you be so kind to review whether I did it properly? :-)

### DIFF
--- a/src/util/updateGameData.js
+++ b/src/util/updateGameData.js
@@ -324,6 +324,8 @@ function Add_Missing_Pokemon() {
 	// Remove missing Diancie and Volcanion
 	delete GameMaster.pokemon['719_0']
 	delete GameMaster.pokemon['721_0']
+	// set proper form name for Armored Mewtwo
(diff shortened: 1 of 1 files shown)


Petap0w:
That should move to add_missing_pokemon function

philcerf:
That doesn't exist yet... so where should I add / call it from?

Also, the pokemon isn't really missing it just has a stupid form name ;-)

Petap0w:
https://github.com/KartulUdus/PoracleJS/blob/2ee5eac4b6e734ac644028fb7caa42d76191b0b7/src/util/updateGameData.js#L304

that's the function to fix those kind of things

philcerf:
@Petap0w done ... please tell whether it's fine by you, and if so, @jfberry please merge :-D

Petap0w:
looks ok

PR #337, opened 2021-02-12
High speed message transmission

jfberry:
Do you want Poracle to **WORK HARDER?**
Do you want to sweat your discord/webhook rate limits?

Well, I have the PULL REQUEST FOR YOU!

Instead of waiting around, Poracle now will despool your messages as fast as it damn well can.
Not only that, but it will use the same bot token to send to multiple destinations at the same time.

Discord's rate limits are per destination so this allows us to get more messages through.

Fine tuning:
* Currently Poracle will send up to 5 message destinations at the same time for discord/telegram
* And up to 10 web hooks concurrently

Are these the right numbers? Well, I don't know. Against my better judgement perhaps I've decided to expose this for you to play with (see below)

Poracle will now log rate limits that it gets told about from telegram, and discord, and webhooks. And it'll backoff for that destination (while still trying others up to the limits above, if it is holding on for all 5 destinations on a discord bot token then communication through that token will hold up).

You can see your Discord / Telegram queues per token in your outbound logs.
You can see rate limits being hit (grep for `Rate limit`)

Not only that, but you can see when users hit their individual rate limit in controller logs.

And fine tuning users rate limits has got more flexible with the new configuration section:

```   "alertLimits": {
        "timingPeriod": 240,
        "dmLimit": 20,
        "channelLimit": 40
    },
```

The timingPeriod is the same as the old alertSecs, but you can now separate out channel and DM limits. Stop individual users hogging your system whilst still letting your channels run free!

This is moved from the discord section as it applies to telegram as well and lots of people probably had the old defaults which were inadequate.

Telegram users? We now send to different destinations concurrently too, and we respect Telegram's rate limit requests also. Something for everyone in this pull request.

** HEALTH WARNING **
Remember tuning is 'in the round'. Change these numbers slowly. Remember that inbound performance is not how long it takes to process an individual alert but how your webhook queue is kept under control; if this clears nicely in between drops from your scanner you are keeping up.  For outbound, lower number of concurrent senders will reduce chance of being rate limited (probably) but to less different destinations.  For reference, old poracle only sent to one destination at a time per bot so 5 is already a big experimental uplift!  

` "tuning": {`

*      `"maxDatabaseConnections": 15`        
* This is the max number of database connections Poracle can have
*      `"concurrentWebhookProcessors": 10`   
* Number of inbound webhooks that can be processed at once. Since each requires a database connection for a query this should be lower than the maxDatabaseConnections
*      `"concurrentDiscordDestinationsPerBot": 5`  
* How many concurrent messages should be queued for sending by the discord.js based bots. Discord rate limits by route as well as a global limit so this will message 5 people at once
*      `"concurrentTelegramDestinationsPerBot": 5` 
* How many concurrent messages should be queued for sending by the Telegram/telegraf based bots. Telegram rate limits by route as well as a global limit so this will message 5 people at once
*     `"concurrentDiscordWebhookConnections": 10` 
* How many concurrent discord webhook destinations to send to at once
`    }`
Note the Poracle senders are route sensitive so the 5 concurrent connections will be to 5 different routes. Messages to the same route will be sent in order, but mainly to give maximum opportunity to not be rate limited for a particular destination set.


jfberry:
Closed, replaced by #358

PR #348, opened 2021-02-13
recude logging

philcerf:
- journald/systemd/console probably doesn’t need to be spammed with verbose
  logging, if everything goes into a file anyway
- webhook logging seems to be purely for debugging, so one can presumably
  disable it per default
- even if enabled, 2 files, each ~1,5 GB, should probably plenty enough

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -26,15 +26,15 @@
         "persistDuplicateCache": false
     },
     "logger": {
-        "consoleLogLevel": "verbose",
+        "consoleLogLevel": "warn",
         "logLevel": "verbose",
         "enableLogs": {
-          "webhooks": true,
+          "webhooks": false,
           "discord": true,
           "telegram": true
         },
         "dailyLogLimit": 7,
-        "webhookLogLimit": 12
+        "webhookLogLimit": 2
     },
     "database": {
         "client": "sqlite",


config/default.json
```diff
@@ -26,15 +26,15 @@
         "persistDuplicateCache": false
     },
     "logger": {
-        "consoleLogLevel": "verbose",
+        "consoleLogLevel": "warn",
```
jfberry, line 29: Think default for this should be verbose. The point is you can override this

config/default.json
```diff
@@ -26,15 +26,15 @@
...
         "logLevel": "verbose",
         "enableLogs": {
-          "webhooks": true,
+          "webhooks": false,
           "discord": true,
           "telegram": true
         },
         "dailyLogLimit": 7,
-        "webhookLogLimit": 12
+        "webhookLogLimit": 2
```
jfberry, line 37: Not sure why a reduction is even needed if the default is to be off (which I guess I can agree with given there could be a lot of disk space used unexpectedly)

philcerf: Well, if someone enables it, he probably still doesn't need 12x 1,5GB of webhook logs, does he? It's just prone to filling the disk.

config/default.json
philcerf: Sure , one can, but most people will stick to the defaults until they run into problems.
And what's the benefit of writing everything twice in verbose? console log will most of the time be caught by journald/syslog and there it's rather not custom to put debug logs in.

And if someone really starts manually an looks at the console it's anyway just for debugging purpose but not normal run.

config/default.json
jfberry: If they are turning it on, they have a reason. Often it is to capture and understand quests or find a specific condition. If they are going to the extent of overriding it they can consider how many hours they get.
Size of the log is dependent on your system of course.

jfberry:
Change to webhook log has been pushed seperately

PR #361, opened 2021-02-24
gym_name is undefined for MAD raid-hooks

fosJoddie:
MAD sends gym name as `name`  both for raids and raid eggs. 

This gets mapped into `gymName` in the handling, but not into `gym_name` which is used for internal logging (and perhaps DTS, who knows)

This oneliner mapes `name` into `gym_name` for complete MAD support

## Description
Fixes "raid at undefined appeared blablabla log messages

## Motivation and Context
Who doesn't love one line PRs?

## How Has This Been Tested?
Arcane Divination 🔮 
image
Here's how this looks like in Visual Studio Code 
## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/raid.js
+++ b/src/controllers/raid.js
@@ -163,6 +163,7 @@ class Raid extends Controller {
 			if (data.name) {
 				data.name = this.escapeJsonString(data.name)
 				data.gymName = data.name
+				data.gym_name = data.name
 			}
 			if (data.gym_name) {
 				data.gym_name = this.escapeJsonString(data.gym_name)


Petap0w:
the idea is to do generally the other way around.
since rdm and mad (and potentially other scanners) are not capable of aligning their var names, we try in Poracle to make one entry available to user (`gymName` - and should be that one used in logs, dts, etc), and add compatibility with scanner vars in code (`if(gym_name) gymName = gym_name` or something like that).
current code isn't using `gym_name` anywhere in logs (or it's a mistake that should be replaced with `gymName`)
users should update their dts to use `gymName` (or if their are MAD then gym_name will always work)

fosJoddie:
🤣  indeed it is fixed there. my local branch was set up to pull from my own repository instead of the main one... 🤦‍♂️

PR #362, opened 2021-02-24
Fix time display issues

jfberry:
There have been some long standing time display issues in Poracle which relate to way that the 'moment' library is used. This could be resolved but this library is broadly discouraged from new use so took the opportunity to fix and move to a more modern javascript library that will increase clarity around time handling going forward.

Because I am based in GMT or BST+1 time variances do not show up very often so this will need testing in other geographies to gain confidence.
Config changes - deprecated: config locale/time
new - locale/time24hr - whether to output display times in 24 hour clock (15:00 vs 03:00 PM)


jfberry:
Answer: moving to luxon library only makes sense for node 14 as full icu installed by default, not willing to take on this complexity when there is another way of fixing the bug without changing library

Issue #391, opened 2021-03-14
Improve the way `maxLimitsBeforeStop` works


kamieniarz:
Right now when user hits the limit they're just stopped like with `!stop` command so in fact users can start them back with `!start` command. The change I'm proposing is a different behaviour of this feature - if user hits the `maxLimitsBeforeStop` limit then notifications would be disabled completly in a way only admin could re-enable with eg `!enable user121212121`  (so in fact hitting limit would trigger sth like `!disable` but after all `!disable` and `!enable` would be admin only commands). This new funcionality could be switchable in config with a string like `"disableUser"` where `true` would make it work like disable/enable and `false` would make it work like stop/start

jfberry:
It was deliberate that users could restart themselves as they could fix their trackings and get going again.  You should be able to see serial abusers through alarm_stats and jump on them that way?

kamieniarz:
I don't want to track what they do cause it's a waste of time. Hitting the limit is still a pretty rare thing so in that case I'd rather explain to user what's wrong and re-enable their notifications instead of letting them do it by themselves. Right now in case they hit the limit the get an info to contact the administrator so they contact me but some of them already figured out all they need to do is `!start`

PR #396, opened 2021-03-16
Update to admin disable capability

jfberry:
Thanks to Len for the main part of this.
Update config to include a roleCheckMode which can be set to disable (do nothing), delete (existing functionality) or disable-user which will mark a user admin disabled.

Also, can set disableOnStop for tracking limits (#391 )


kamieniarz:
This PR is just awesome - good job!

Things that could be added:
1. Add commands `!userlist enabled` and `!userlist disabled`
2. Add a notification to `dmLogChannelID` that would poke us (admins) if user hits `maxLimitsBeforeStop`

PR #397, opened 2021-03-18
March Quality of Life #7

jfberry:
Running out of these small changes, so we must be on target for a push to master at the end of the month.
1. `/userlist` extended for enabled/disabled as parameters to filter
2. Log channel gets a message when users get stopped (hit their final rate limit)
3. handlebars available as Eng as well as Alt (so pokemonNameEng for example)
4. quest details available as Eng too (questStringEng and rewardStringEng)
5. new 'hidden' config option config.general.ignoreMADQuestString - now there is a PR from Joddie to add the extra data to the poracle quest from MAD we can generate this content internally to match the format of the string we generate for RDM (this simplifies translation).  Opt in for this if you are testing the MAD PR -- it will become the default later

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -210,7 +210,7 @@
         "#FF8000"
     ],
       "dmLogChannelID": "",                           // an ID of a channel to use to log all poracle commands to (to watch users)
-      "dmLogChannelDeletionTime": 0,                  // time to clean up after (seconds) - 0 do not delete
+      "dmLogChannelDeletionTime": 0,                  // time to clean up after (minutes) - 0 do not delete
       "messageDeleteDelay": 0,                        // extra time to add on to a message 'clean' to leave behind (in ms)
     // unrecognisedCommandMessage - message to reply to users who send an unrecognised command in DM
       "unrecognisedCommandMessage": "",
--- a/src/app.js
+++ b/src/app.js
@@ -324,6 +324,7 @@ async function processMessages(msgs) {
 		const rate = rateChecker.validateMessage(destinationId, destinationType)
 
 		let queueMessage
+		let logMessage = null
 
 		if (!rate.passMessage) {
 			if (rate.justBreached) {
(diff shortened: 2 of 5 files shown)


kamieniarz:
Here are my thoughts. Some of them are about this PR but some of them are about PR 396 also:

1. Info about stopped alerts should poke me as an admin so I would be notified if rate limit was reached
2. User can do `!unregister` and register again in dedicated channel to skip rate limit without enabling him back
3. Actually I think `!unregister` should fully remove user from database cause right now it works like disable or like removing access role
4. Users who are already in database and get re-enabled won't get welcome message - I think they should get same info as new users which is welcome message + help (or maybe sth else?). It's always an info that informs a users that he can now use notifications.

jfberry:
> 1. Info about stopped alerts should poke me as an admin so I would be notified if rate limit was reached

Via DM, or just ping you in the channel? As I was writing the alert I thought this might evolve into a more generalised "draw something to the attention of the admin" capability.  This might be one to hold over as there might be some general settings about how to alert and who to alert which might align more closely with my delegated administration changes I have planned 

> 2. User can do `!unregister` and register again in dedicated channel to skip rate limit without enabling him back
> 3. Actually I think `!unregister` should fully remove user from database cause right now it works like disable or like removing access role

Actually (3) is already how it works - unregister does cause a delete; and I think that is right as it is a user self selected unregister.  There is a question here actually about the future reconcilliation capability as if they retain the role they would then get added back. Hmm.
I'm not sure what you can do about (2), frankly.  If you have a user that does this then perhaps you should ban them entirely.

> 4. Users who are already in database and get re-enabled won't get welcome message - I think they should get same info as new users which is welcome message + help (or maybe sth else?). It's always an info that informs a users that he can now use notifications.

Users who !poracle do get the greeting, have added back to role change for disabled users becoming enabled (

kamieniarz:
Ok looks like all is clear now and thing work properly, thanks. About ping - just in log channel. I'm not following all logs so without any kind of ping I'll just miss the fact someone was disabled

PR #407, opened 2021-04-04
Implemented support of clean for discord webhook worker

jfberry:
Thanks Len for pointing out that discord webhooks now support deletes (perhaps they always did and I missed it).
So this now allows support of cleaning of messages sent by discord webhooks!


lenisko:
Changes looks good. I will test it today and leave feedback :+1: thanks for fast reaction!

lenisko:
Running it on prod from 2 hours and no problem. Cleanup after restarting PoracleJS tested as well.

PR #423, opened 2021-04-23
Option to ignore raid hour eggs/raids

jfberry:
New option in general:
```
      "ignoreLongRaids": false                // ignore raids > 47m long to save raid hour spam
```


plinytheelder:
worked like a charm.

jfberry:
Super, I may change the config option naming slightly but will incorporate something along these lines very soon!

plinytheelder:
for sure. Maybe change it to `ignoreEventRaids` since these things pop up every so often.

jfberry:
or 'ignoreLongRaids' ... or... have it as a maximum value. Not quite sure what to do - but agree really helps keep down channel/bot spam

PR #451, opened 2021-05-21
fix lint

KartulUdus:
## Description

fixed linting errors accumulated over time 
## Motivation and Context

javascript is a loosley typed language and loose equal should be avoided if at all possible to avoid type coercion 
## How Has This Been Tested?

not properly tested (yet)
## Types of changes

## Checklist:


src/lib/poracleMessage/commands/egg.js
```diff
@@ -108,7 +107,6 @@ exports.run = async (client, msg, args, options) => {
 									...toInsert,
 									uid: existing.uid,
 								})
-								insert.splice(i, 1)
```
jfberry, line 111: I see the change from updates to deletes & inserts, which works on the database side but the user feedback will be duplicated (ie have an entry as 'New' and another as 'updated')

src/lib/discord/commando/index.js
```diff
@@ -60,6 +62,7 @@ class DiscordCommando {
 			this.client.logs = this.logs
 			this.client.dts = this.dts
 			this.client.re = this.re
+			this.client.updatedDiff = this.updatedDiff()
```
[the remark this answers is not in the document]
KartulUdus: doesn't work tho. 

Any idea why this or down by L71 doesn't expose it in client.updatedDIff the command itself?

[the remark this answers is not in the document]
jfberry: 'client' inside the commands actually a poracleDiscordState or poracleTelegramState which provides a generalised interface, probably needs to be carried through to there - on mobile now so can't look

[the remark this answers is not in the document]
KartulUdus: got that import resolved. 

telegram didn't use objectDiff as far as I saw

src/lib/poracleMessage/commands/egg.js
KartulUdus: worked around it in abca4fe. 
I still believe that a delete + insert is better than an update call for each mon

src/lib/discord/commando/index.js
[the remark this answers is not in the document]
jfberry: The commands are the same for telegram and discord.  Discord capability is encapsulated in poracleDiscordState, poracleDiscordMessage and poracleDiscordUtil; analogous classes exist for telegram to provide compatible capability so that by the time commands are running they don't care about the communication mechanism.
So poracleTelegramState will need similar massaging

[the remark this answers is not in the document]
KartulUdus: right... so 
https://github.com/KartulUdus/PoracleJS
already carries over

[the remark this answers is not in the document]
jfberry: Right - so poracleTelegramState (which is created in there) needs updatedDiff from somewhere

src/lib/poracleMessage/commands/egg.js
```diff
@@ -133,12 +132,14 @@ exports.run = async (client, msg, args, options) => {
 				})
 			}
 
-			if (insert.length) {
```
jfberry, line 136: This fix was because insertQuery fails with an empty array, and if all the additions are eliminated (because they go into the unchanged list), then it was causing an exception. I think the same condition exists in your revised code, but now check should be ... `if (insert.length || updates.length)

KartulUdus: why aren't we using upsert for tracks any more?

jfberry: Too many columns for mysql to support the required index

jfberry: was a depressing time, mysql and mariadb have different limits in this area too :-(

KartulUdus: I don't know why db methods are duplicated between controller.js and query.js but I've added the crutch in a single place instead of each command

jfberry: Short answer - start of some refactoring not yet complete.

Longer answer - controller.js being the base class for all of the alarm controllers was quite large and had many more dependencies than were really needed by simple database work.  So first part of refactoring was to move out the database access which is used more broadly (outside of alarms in many places).
It turns out that the commands use some other things which aren't queries which had to be moved across also.
But these should really be split out, and then controller.js pointed to query so that db access is through one place.

Crutch makes sense.

PR #453, opened 2021-05-26
bugfix - wait too long after error 429 on discord webhook

jfberry:
Fix

### DIFF
--- a/src/lib/discord/discordWebhookWorker.js
+++ b/src/lib/discord/discordWebhookWorker.js
@@ -63,7 +63,7 @@ class DiscordWebhookWorker {
 					if (!res.headers.via) {
 						this.logs.discord.error(`${senderId} WEBHOOK 429 Rate limit [Discord Webhook] TELL @JABES ON DISCORD THIS COULD BE FROM CLOUDFLARE: ${retryAfterMs}`)
 					}
-					await this.sleep(retryAfterMs + Math.random() * 5000)
+					await this.sleep(+retryAfterMs + Math.random() * 5000)
 					retry = true
 					retryCount++
 				}


src/lib/discord/discordWebhookWorker.js
```diff
@@ -63,7 +63,7 @@ class DiscordWebhookWorker {
 					if (!res.headers.via) {
 						this.logs.discord.error(`${senderId} WEBHOOK 429 Rate limit [Discord Webhook] TELL @JABES ON DISCORD THIS COULD BE FROM CLOUDFLARE: ${retryAfterMs}`)
 					}
-					await this.sleep(retryAfterMs + Math.random() * 5000)
+					await this.sleep(+retryAfterMs + Math.random() * 5000)
```
KartulUdus, line 66: won't fix the problem, you're expecting parseInt(undefined) * 5000 is better than undefined * 5000
both cases parse to NaN

should be something like `await this.sleep(retryAfterMs? retryAfterMs + Math.random() * 5000) : 5000`

jfberry: That would be safer, but retryAfterMs is present and is a string in the case was solving for

KartulUdus: weird, res.headers['retry-after'] was not undefined but some string, what string?

jfberry: 1059 or some such... as expected. So actually ended up waiting 10593000 ms or something like that

KartulUdus: what I mean -> 
image

jfberry: and the other warning there is because I read that if cloudflare sends the 429 not discord it is in seconds not ms (spec says seconds but discord have it wrong).  But I have never seen one ;-)

KartulUdus: +String + otherMath would only work if it's an _empty_ string

+var is short for parseInt(var)

PR #454, opened 2021-05-29
Add areas to weatherchange

jfberry:


config/default.json
```diff
@@ -182,7 +182,7 @@
     //    deny: users cannot track for everything and must track against individual pokemon
```
KartulUdus, line 182: does default.json no longer work because of comments in json file?

jfberry: No all fine. Node config supports json comments

jfberry:
will include in future PR rather than uplifting this

PR #469, opened 2021-06-21
Support for !quest form and !quest candy

jfberry:
**DATABASE MIGRATION**

Quests support forms and candy rewards!

1. `!quest slowpoke formgalarian`
2. `!quest candy`
3. `!quest candybulbasaur`

Also introduced change to query to support minimum amounts of mega energy and candy but not added to the quest command


Kneckter:
This PR has been working for me so far. Mainly tested the candy tracking.

PR #470, opened 2021-06-21
Future event forecasting

jfberry:
Don't you just hate it when a spawn changes after notification?
Thanks to Malte's event list Poracle can now predict if a spawn or quest is going to change and provide a warning.

Nothing to activate, it just works (or at least it should)

Include something like this in your dts:
```{{#if futureEvent}}There is an event ({{futureEventName}}) {{#eq futureEventTrigger 'start'}}starting{{else}}ending{{/eq}} at {{futureEventTime}}.{{/if}}```


bbdoc:
Wouldn't it be better to mimic the weather warnings and also have a specific event showing all impacted pokemon at the time of the event change ?

jfberry:
Perhaps it would be better, but substantially more complicated so can come later - I will add to wishlist voting after this one is merged

plinytheelder:
If we’re scraping info off his event repo, why not:

- some !events command that can spit out active event info to a discord channel. Maybe even user.
- One that posts new events.

jfberry:
When combined with the new !info PR I can imagine an `!info event` to understand current event at very least.
I'm not sure about event posting... but perhaps poracle could consider the events like hooks eventually

jfberry:
Added capability for futureEvent in quests

PR #472, opened 2021-06-28
Update Util JSON

Kneckter:
Ran the update script to pull in missing entries in the JSON files. Also updated the util.json file for capital letters, CP multipliers, and the gen 8 ID range.


jfberry:
Thanks... I will need to verify the text changes to util.json to see how they impact translations before merging. If these could be in a separate PR then could get the rest in straight away?

Kneckter:
There you go, I split them apart =]

PR #473, opened 2021-06-28
Updated util.json

Kneckter:
Updated the util.json file for capital letters, CP multipliers, and the gen 8 ID range. Split from PR #472


jfberry:
Thanks, will check in the code how these text items are used when I get a moment :-)

jfberry:
Have brought in cpMultipliers in !info commit;  haven't created a space in my head to think about implications for changing weather.
You know you can override this in your custom.en.json file if you want (or whichever language you are using)

PR #474, opened 2021-07-02
Support for Obhem to calculate PVP stats

jfberry:
*** DATABASE MIGRATION ***

This PR adds support for Poracle to use Mygod's new Obhem PVP library. You need to `npm install`

This has enabled:

* Support for little cup pvp tracking
* Removing the need for PoraclePvpHelper for MAD installations

At this stage we are testing whether there is a particular impact on memory or performance.  It looks like 60mb per worker thread (300mb on a default instance) with an acceptable performance impact after it has warmed up.
 
By default, debug log entries that show the PVP calcs from webhooks and obhem are logged.  You can have poracle ignore webhooks and process according to obhem using a new configuration option under PVP which can be 'internal' (use obhem) or 'webhook'.  Obhem can consider mega evolutions, and also multiple level caps simultaneously if you want to.

```
    "pvp": {
        "dataSource": "internal",
         "levelCaps": [50],                            // level caps to be included in internal rank calculations - alternative could be [50, 51]
         "includeMegaEvolution": false,                // whether to include mega evolutions in rank calculations

```

Log entry:
```
2021-07-07 18:25:25 $2 verbose: 4621680556270937899: Darumaka{953/28.89} appeared at [51.277,1.088] areas (xxxx,yyyy) and 0 humans cared. (17.064464/0.332367 ms)
```
The timing in brackets is the total processing time for the alert matching (as before) with the PVP calculation broken out [it would be included].  The log entry now also includes cp/iv of the mon to make any diagnosis easier.

Little cup mons can be tracked with 'little' and 'littlecp' and your dts can be upgraded to display little league in line with ultra/great. Example will follow.  Little cup only works with obhem calculation -- ie set dataSource to internal.

** New DTS **

You can continue to use the old convoluted DTS, or use some new DTS magic which does all the filtering and name
expansions for you

here is my test DTS:

```
{{#if pvpLittle}}**Little league:**\n
{{#each pvpLittle}}- {{this.fullName}} #{{this.rank}} @{{this.cp}}CP (Lvl. {{this.level}})\n
{{/each}}{{/if}}{{#if pvpGreat}}**Great league:**\n
{{#each pvpGreat}} - {{this.fullName}} #{{this.rank}} @{{this.cp}}CP (Lvl. {{this.level}})\n
{{/each}}{{/if}}{{#if pvpUltra}}**Ultra League:**\n
{{#each pvpUltra}} - {{this.fullName}} #{{this.rank}} @{{this.cp}}CP (Lvl. {{this.level}})\n
{{/each}}{{/if}}
```

Note that it automatically filters by your display filters and you have direct access to the following things in the each loop:
* fullName - a full name including form (and mega)
* fullNameEng - english version
* rank
* cap (the cap of this entry, eg 50 or 51)
* formId
* pokemonId
* evolution (whether this is a mega or not)
* cp
* percentage (2dp)
* monsterName - just the pokemon name
* formName - just the form name
* baseStats - for quick use of calculateCp

Feedback?
I'd appreciate any testing from interested MAD users who would like to try sending encounters directly; or from anyone else who looks at memory use or performance before or after (the timing line pre scan will show the obhem impact - to be fair to it look after a little while so it can get warmed up) or to track 

---
If you ran an earlier version of this PVP pull request, these commands remove the old columns (it will lose your old little cup tracks too)
```
delete from monsters where little_league_ranking < 4096;
alter table monsters drop little_league_ranking, drop little_league_ranking_min_cp, drop little_league_ranking_highest, drop great_league_ranking_highest, drop ultra_league_ranking_highest;
delete from migrations where name='v09_littlecup.js';
```


FimoJones:
This is working nicely and it was a painless transition. I have not embraced your new DTS logic yet @jfberry but I am sure it will also transition nicely.

PR #477, opened 2021-07-06
Auto reload DTS file on change

jfberry:
Initial version - doesn't reload for weather controller or fully pass lint yet


jfberry:
Merged into QOL PR

PR #486, opened 2021-07-11
initConfig.json: Use correct regex for telegram

reedy:


jfberry:
Unfortunately needs to be rebased because of your other pr. how frustrating

PR #489, opened 2021-07-11
Remove space before commands in help

reedy:
Screenshot 2021-07-11 at 14 56 41


reedy:
There's a few more of these where it's inconsistent

PR #495, opened 2021-07-19
July QOL2 - Sane Emoji translation

jfberry:
1. Current strategy with emojis is to use translation to change to custom emojis. But this has the disadvantage of not differentiating between discord & telegram giving rise to some users not being able to use this.  Translation still works with this version, but a new more sane emoji logic has also been introduced.

The `util.json` now contains a mapping from standard strings like `type-dark`, and `weather-cloudy` to emojis.  These can be overridden in an `emoji.json` in your config folder.  It can contain different mappings for telegram and discord.  Emojis can be mapped to any string.

My test looks like this:
```
{
  "discord": {
    "weather-sunny": "SUNNY",
    "weather-partly-cloudy": "PARTLY CLOUDY",
    "grunt-unknown": "WHAT?"
  },
  "telegram": {

  }
}
```

2. Stop on migration errors -- Poracle will now stop if there is a migration error rather than continuing (--force command line arg will make it continue)

3. DTS auto re-load on file change

4. gym tracked text now shows if slot changes are active

5. in aid of DTS debugging if DTS generation fails then poracle will still post a generic message containing reference information to allow a quick solution

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,39 @@
+name: publish
+
+on: [push]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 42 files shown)


jfberry:
Brought up to date with PVP patch, will be next to be merged

PR #502, opened 2021-08-08
Add monster information to !info

jfberry:
Adding monster (dex) lookup to !info command.  Output format quite rough at the moment, intended to give useful game data. 
At the moment it contains forms poracle supports, Hundo CP at various levels, type information, weathers boosted, weak against and strong against.

image

image


plinytheelder:
By the way, Level 15 is what we get from Research. Might also be worth adding.

PR #503, opened 2021-08-09
Discord command security

jfberry:
**DATABASE MIGRATION**

Have you ever wanted to restrict commands to users with particular roles?

All commands and all parameters for the track command now supported.

```
    // commandSecurity - list commands and users/roles that can access certain capabilities. If not specified no
    // security is enforced for a command
    // eg:
    //   "commandSecurity": {
    //      "monster": [ "userid", "roleid" ],
    //      "pvp": [ "roleid" ]
    //   }
      "commandSecurity": {

      },
```

Valid command types are monster, lure, invasion etc - in addition pvp to restrict access to pvp tracking

In addition the !track command will now return errors if invalid parameters are used


kamieniarz:
I'm just testing this PR and I see some issues. To make it all clear lets say role A has access to `!track` and `!untrack` and role B doesn't have the access. Both roles are listed in config in `userRole` section. Below are my observations:
1. When I have role B `!track` and `!untrack` commands return error - that's good, it works
2. When I have role B and give myself role A every command works - that's also good
3. When I take away role A (role B stays) poracle unregisters me - not cool, access should stay since there's still a role that is listed in `userRole`. It doesn't work like that in other direction, eg. when I have role A and B and take away role B - role A still works without unregistering and I think that's how it should generally work.
4. When I had access to role A I could use `!track` command as I wanted. When I lost the role I was unregistered. When I gave myself B role I was enabled with all settings from A role, so in fact I was tracking mons with role that has `!track` disabled - that's the worst case in this PR.

jfberry:
> I'm just testing this PR and I see some issues. 
The PR is intended to restrict access to commands, not validate continuously what you have access to.
So if you set trackings with a role in place and lose the role your trackings will not be revoked.  It does not re-evaluate your capabilities and I have no intention of doing that.

For your third point that sounds like a general unregistration issue with two roles and losing one? (this PR does not change anything about registration/unregistration). I

kamieniarz:
> > I'm just testing this PR and I see some issues.
> > The PR is intended to restrict access to commands, not validate continuously what you have access to.
> > So if you set trackings with a role in place and lose the role your trackings will not be revoked.  It does not re-evaluate your capabilities and I have no intention of doing that.
> 
> For your third point that sounds like a general unregistration issue with two roles and losing one? (this PR does not change anything about registration/unregistration). I

That's proly it. Weird thing in this case is that it's an issue only when removing role A, not the other way. I tested it with few roles with same permissions and it always looked the same. It's ofc not a big deal after all, at least not as big as 4th point

jfberry:
I just checked point (3) and everything worked as it should - users only got unregistered when they lost all roles (that's nothing to do with this PR, but I wanted to check it was right). So understanding that separately to see if there is a problem would be useful.

Point (4) is as designed. Of course if you have delete user instead of disable user you don't get to keep your existing tracks after you lose access.

kamieniarz:
Ok I just reverted and made some tests regarding 3rd point - same issue so I'll investigate later.

About 4th one - that's a crucial one in my case cause I want to keep `"roleCheckMode": "disable-user"`. Also I have users who mostly during CD switch from basic access (without mons) to full access (with mons) and the way this PR works now simply won't work for me with `disable-user`. Maybe some checks that would remove things? Eg. if current role is in `commandSecurity` and not in `track` then remove from tracking all mon related, if not in `invasion` then remove all invasion related, etc etc. I'm just thinking fast... Anyway despite that all seems fine to me

jfberry:
DM me about the 3rd issue so I can try and recreate (my tests showed it worked fine so there is obviously some config difference between our tests)

> About 4th one - that's a crucial one in my case cause I want to keep `"roleCheckMode": "disable-user"`. Also I have users who mostly during CD switch from basic access (without mons) to full access (with mons) and the way this PR works now simply won't work for me with `disable-user`. Maybe some checks that would remove things? Eg. if current role is in `commandSecurity` and not in `track` then remove from tracking all mon related, if not in `invasion` then remove all invasion related, etc etc. I'm just thinking fast... Anyway despite that all seems fine to me

A delete of all monster tracks would on the face of it seem quite straight forward.  But under what conditions would it be performed?
Say you have role A for mons, and B for raids
User has role A & B. On drop of A we delete the mon tracks.

Same rule applies, user has just A. On drop of A we delete the mon tracks.  Except you want disable-user to work.

I just don't think the logic is compatible.

kamieniarz:
This PR is not for me then. I'll stay with 2 instances 🤷‍♂️ Anyway looks like PR works as intended ;)
Maybe solution I need would require another PR or just leave it like that untill more ppl ask for it

TurtIeSocks (approved):
Everything is working great!

As previously discussed, maybe a future PR to add filtering to the controllers to stop sending alerts for users that used to have certain permissions. Thanks @jfberry :)

jfberry:
Added immediate blocking of alerts when users lose role. 
Alert type added to new blocked_alert column which is checked by the controllers.
In addition, users can always untrack and tracked tells you if you are blocked from something.

I think this solves your problem @kamieniarz

PR #520, opened 2021-08-23
Discord Status/Activity Expansion

TurtIeSocks:
## Description

- Cleanup code for custom Discord activity on primary worker
- Allowing customizing the status and activity of the additional Discord workers

## Motivation and Context

- Extra customization for map admins

## How Has This Been Tested?

Works locally

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/discord/discordWorker.js
+++ b/src/lib/discord/discordWorker.js
@@ -3,7 +3,7 @@ const fsp = require('fs').promises
 const NodeCache = require('node-cache')
 const FairPromiseQueue = require('../FairPromiseQueue')
 
-const noop = () => {}
+const noop = () => { }
 
 class Worker {
 	constructor(token, id, config, logs, rehydrateTimeouts = false) {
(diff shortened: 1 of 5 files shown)


src/lib/discord/discordWorker.js
```diff
@@ -63,7 +63,10 @@ class Worker {
 		try {
 			await this.setListeners()
 			await this.client.login(this.token)
-			await this.client.user.setStatus('invisible')
+			if (this.token !== this.config.discord.token[0]) {
```
jfberry, line 66: Does this work here? I thought activity had to be set in ready?
If it does work here we should potentially do all the sets here and remove the set from ready
It would be better to pass in an object 'discordStatus' with visibility and activity through the constructor then the caller could set these as it loops through creating the objects and this wouldn't need to understand about any other tokens.

jfberry:
You got rid of my enforced poracle advertising  :-). I don't mind though.
Incidentally, I doubt you actually need any secondary bots any more. The new discord sender is much more efficient at managing routes and I don't know anyone (that has actually tried to reduce their bots) that needs more than the main one - which is obviously less confusing to users.

src/lib/discord/discordWorker.js
TurtIeSocks: It was inconsistent. Sometimes I would start it and both the main bot and the worker would have the activity set here and sometimes only the worker bot would have the activity set here, so I added the check. I'll test to see if removing it from the ready.js would make it consistently be set from here.

PR #521, opened 2021-08-24
Non-Invasive Chuck Support

TurtIeSocks:
## Description

- Adds the bare minimum Chuck PVP support
- Tracks level caps in DTS
- Attempted to make it the least invasive as possible

## Motivation and Context

I would like it to work with my parser, even if I'm the only one using it.

## How Has This Been Tested?

This has been tested with RDM and Chuck and still works fine with each.

Screen Shot 2021-08-23 at 8 06 48 PM

## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/monster.js
+++ b/src/controllers/monster.js
@@ -311,6 +311,7 @@ class Monster extends Controller {
 								form: stats.form || 0,
 								level: stats.level,
 								cp: stats.cp,
+								cap: stats.cap,
 							}
 							data.pvpEvolutionData[stats.pokemon] = {
 								...data.pvpEvolutionData[stats.pokemon],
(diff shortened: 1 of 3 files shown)


jfberry:
This is fine I think.
Is the pokemonLevel handlebars function needed when there is a 'levelWithCap' expansion in the new dts for the leagues? (or is this mainly for compatibility with your previous implementation?)

TurtIeSocks:
> This is fine I think.
> Is the pokemonLevel handlebars function needed when there is a 'levelWithCap' expansion in the new dts for the leagues? (or is this mainly for compatibility with your previous implementation?)

Hadn't realized it was there actually, confirmed it's workin gas anticipated and have removed the helper.

PR #525, opened 2021-08-25
update de.json and add DE translation for !info

ReuschelCGN:
## Types of changes

## Checklist:


config/locale/de.json
```diff
@@ -226,7 +235,10 @@
     "Fire":"Feuer",
     "Flying":"Flug",
     "fog":"Nebel",
-    "form":"form",
+    "Forecast":"Vorhersage",
+    "form":"Form",
+    "formalola":"alolaform",
```
jfberry, line 240: I think this may break tracking

config/locale/de.json
```diff
@@ -423,6 +450,7 @@
     "Silver Pinap Berry":"Silberne Sananabeere",
     "Sinnoh Stone":"Sinnoh-Stein",
     "Sky":"Zenit",
+    "slot_changes":"Platzänderung",
```
jfberry, line 453: this will need to be without _

ReuschelCGN: yeah your right -> fixed!

config/locale/de.json
jfberry: ie formalola --> the regex expects form to be the parameter name, and then looks up the form type afterwards. 
ALthough maybe this means you can type alolaform and have it translated back into formalola which would work...

It makes my head hurt ;-)

config/locale/de.json
ReuschelCGN: ok i removed it! now it should be fine

config/locale/de.json
ReuschelCGN: ok i removed it! now it should be fine

PR #542, opened 2021-09-01
Ditto disguise name handlebars

jfberry:
{{disguisePokemonName}}

{{name}}{{#if disguisePokemonName}} (disguised as {{disguisePokemonName}}){{/if}}

This requires display_pokemon_id in hooks which RDM sends but MAD currently doesn't. I am unable to test.


src/controllers/monster.js
```diff
@@ -425,6 +425,16 @@ class Monster extends Controller {
 				return []
 			}
 
+			if (data.display_pokemon_id) {
```
fosJoddie, line 428: trolly webhooks might send pokemon_id as display_pokemon_id, a simple `!=` would make sure it never happens

jfberry: agreed, added

fosJoddie (approved):

PR #567, opened 2021-09-27
Fixed enable/disable command bugs 

hokiepokedad2:
There were errors execute the update query when executing either the enable or disable commands. The 'id' values weren't getting passed in. 

## Description
Included the 'id' value in their respective poracleMessage commands. Also fixed enumeration of mentions in the enable command. 

## Motivation and Context
Needed commands to work. 

## How Has This Been Tested?
Allow rate limiter function to stop/disable the user. Attempt to perform enable. 

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x ] My code follows the code style of this project.

### DIFF
--- a/src/lib/poracleMessage/commands/disable.js
+++ b/src/lib/poracleMessage/commands/disable.js
@@ -22,7 +22,7 @@ exports.run = async (client, msg, args, options) => {
 		for (const id of targets) {
 			client.log.info(`Disable ${id}`)
 
-			await client.query.updateQuery('humans', { admin_disable: 1, disabled_date: null }, { id })
+			await client.query.updateQuery('humans', { admin_disable: 1, disabled_date: null }, { id: id })
 		}
 		await msg.react('✅')
 	} catch (err) {
--- a/src/lib/poracleMessage/commands/disable.js
+++ b/src/lib/poracleMessage/commands/disable.js
@@ -22,7 +22,7 @@ exports.run = async (client, msg, args, options) => {
 		for (const id of targets) {
 			client.log.info(`Disable ${id}`)
 
-			await client.query.updateQuery('humans', { admin_disable: 1, disabled_date: null }, { id: id })
+			await client.query.updateQuery('humans', { admin_disable: 1, disabled_date: null }, { id })
 		}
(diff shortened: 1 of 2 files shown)


src/lib/poracleMessage/commands/disable.js
```diff
@@ -22,7 +22,7 @@ exports.run = async (client, msg, args, options) => {
 		for (const id of targets) {
 			client.log.info(`Disable ${id}`)
 
-			await client.query.updateQuery('humans', { admin_disable: 1, disabled_date: null }, { id })
+			await client.query.updateQuery('humans', { admin_disable: 1, disabled_date: null }, { id: id })
```
jfberry, line 25: Doesn't this just get changed back by lint? It certainly would by webstorm. It's equivalent to what was there before

src/lib/poracleMessage/commands/enable.js
```diff
@@ -17,12 +17,12 @@ exports.run = async (client, msg, args, options) => {
 		// Make list of ids
 		const mentions = msg.getMentions()
 		const targets = mentions.map((x) => x.id)
-		targets.push(args.filter((x) => parseInt(x, 10)))
+		targets.push(...(args.filter((x) => parseInt(x, 10))))
```
jfberry, line 20: This is absolutely a bug (this change mirrors disable)

src/lib/poracleMessage/commands/disable.js
hokiepokedad2: You're right. This was test code to attempt to isolate what was going on. This has been reverted.

src/lib/poracleMessage/commands/enable.js
hokiepokedad2: Do you mean you agree that it was a bug or the fix is a bug?

jfberry: I agree that your change fixes the bug! Sorry if I was unclear

hokiepokedad2: Ah okay, no problem!

PR #604, opened 2021-11-04
translation for !poracle-clean & de.json

ReuschelCGN:
!poracle-clean command messages "to cleanup messages in discord" will be translated.

Keys to translate in other languages:
    "Cleaning finished":"   ",
    "Will start cleaning up to {0} messages back - do not re-run until finished":"   ",

Tested in my local bot!


src/lib/discord/commando/commands/poracle-clean.js
```diff
@@ -8,15 +8,15 @@ exports.run = async (client, msg) => {
 		const limit = 100
 		const messages = await msg.channel.messages.fetch({ limit })
 
-		const startMessage = await msg.reply(`Will start cleaning up to ${limit} messages back - do not re-run until finished`)
+		const startMessage = await msg.reply(`${client.translator.translate('Will start cleaning up to')} ${limit} ${client.translator.translate('messages back - do not re-run until finished')}`)
```
jfberry, line 11: Thank you for this!
Did you verify that this works (this is a pure discord command rather than having all the poracle infrastructure available so couldn't tell you off the top of my head)
If translateFormat is available here then 'Will start cleaning up to {0} messages etc' might be better than having two strings

PR #617, opened 2021-11-20
Add a config option to telegram dts to send text after location

Tabbomat:
## Description

For Telegram, a notification usually consists of a Sticker, followed by text, then the location of the object. This PR adds an option `text_last` to telegram dts templates to send the text after the location.

## Motivation and Context

When the text is sent last, you can quickly check what that object is in the chat overview. However, when the location is sent last, you only get a "Location" in the chat overview, and you have to tap on the channel/bot to see what was sent. This saves about 2 seconds for the users.

## How Has This Been Tested?

I have tested this on my personal instance serving only telegram users.

## Types of changes

This does not change anything for discord, and does not change the default Telegram behaviour. This is an opt-in change, existing dts templates do not need to be updated to contain the option.

## Checklist:

The default dts.json has been updated to show the usage.

I am aware that the if block is duplicated. Since `messageIds` must be updated and my javascript knowledge is very limited, I could not think of a better way to do this.


jfberry:
I am happy to support this. Do you think it would be better to have an order where you could do:
send_order: ['sticker','location','text'] or something like that?

Tabbomat:
Basically a for loop over `send_order`, containing a switch block?

What shoud happen when you set `send_order: ["sticker", "text"]` and `location: true`? Don't send location, send location last or some error? In the spirit of backwards compatibility I don't want to retire the option `location`, while at the same time allowing incomplete `send_order` lists (if you only want sticker and location for some reason).

jfberry:
Yes, a loop with a switch would be fine.
It should default to the current order if none specified (ie you can do `for (const part of send_order || ['sticker','text','location])`

If a bit is missing then don't send it - easier that way - it's a misconfiguration

jfberry:
Are you going to update this? I will merge if you do.
Alternatively if you are not, I make the change and add it to my christmas PR

Tabbomat:
Sorry for the long delay, due to work I had to put this on the back burner and then forgot about it.

I tried making this look nicer. For example only one try catch block containing a switch statement within the for loop, but then there would still be ifs required for checking if the message contains the sticker/photo/location. Also the logging statements were not uniform across the 4 message types.

There is some input sanitation, and the default behaviour should be unchanged.

jfberry:
Thanks for this - there are a few changes I would make, but looks great. I will take it into my december QOL PR and make the changes there rather than going backwards or forward. Will be merged soon :-)

PR #619, opened 2021-11-30
PVP Level Cap Filtering

TurtIeSocks:
## Description

- Cap migration
- Cap command line inputs
- Vague cap filtering, only works ~50% of the time
- Uncapped Pokemon always go through

Not 100% convinced this is the way to implement this. It might be more useful to just do a boolean XL column and filter out all Pokes that are above or below L40. Since in its current state, most Pokemon are still "uncapped" and get through the filter

This is also just not working 100% yet as mentioned above, I fear that there may be a larger rewrite necessary to get this to work more consistently.

I would not recommend testing yet in a production env since the migration may change before final.

## Motivation and Context

- Allows users to better filter out the desired PVP mons

## How Has This Been Tested?

Locally. Not 100% working yet.

## Types of changes

## Checklist:


jfberry:
Replaced by #646

PR #629, opened 2021-12-20
!broadcast to users based on area or location; !poracle-test to force test messages

jfberry:
1. Broadcast feature

`config/broadcast.json` defines messages that can be broadcast to users. An example broadcast.json file is provided.  Discord messages can be embeds

`!broadcast area:New_York donate`
`!broadcast area:New_York area:California donate`
`!broadcast 51.1,1.1 d10000 donate`

You can test a particular message id `!broadcast test donate` which will queue the message to you.

2. Testing dts

You can define some testdata in `config/testdata.json` which you can then use to test poracle dts.  An example is provided.  You can have the message pushed back to you choosing a particular template.  If the location is set to keep, the longitude/latitude in the message is used, otherwise it is updated with the current location from your humans entry.  The times in the message are brought up to current time always.

`!poracle-test raid level5`
`!poracle-test pokemon boring templatepvp`


jepke:
🙏

PR #634, opened 2021-12-26
Telegram allow send_order

jfberry:
Update of PR #617 

Telegram - allow adding a send_order -- eg:

`      "send_order": ["sticker","location","text"]`

to control the order of the individual messages being sent by Poracle in an alert


jfberry:
@Tabbomat sorry for the delay; I think this is right (and is a small modification on your PR) - ready to merge if you say

PR #652, opened 2022-01-11
Add disabledHooks to apiConfig

Petap0w:
Fetches `config > general > disablePokemon / disableRaid / disablePokestop ...` to return in apiConfig call an array `disabledHooks = [ 'weather', 'nest' ]` of hooks that are set to `disableXXXXX = true`

### DIFF
--- a/src/routes/apiConfig.js
+++ b/src/routes/apiConfig.js
@@ -32,7 +32,8 @@ module.exports = async (fastify, options, next) => {
 			},
 			maxDistance: fastify.config.tracking.maxDistance,
 			everythingFlagPermissions: fastify.config.tracking.everythingFlagPermissions,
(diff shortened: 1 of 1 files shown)


src/routes/apiConfig.js
```diff
@@ -32,7 +32,9 @@ module.exports = async (fastify, options, next) => {
 			},
 			maxDistance: fastify.config.tracking.maxDistance,
 			everythingFlagPermissions: fastify.config.tracking.everythingFlagPermissions,
-
+			disabledHooks: ["Pokemon", "Raid", "Pokestop", "Invasion", "Lure", "Quest", "Weather", "Nest", "Gym"].flatMap((hookType) => (
```
jfberry, line 35: I had to bend my head around this expression.

Does this achieve the same thing in a simpler form?

`["Pokemon", "Raid", "Pokestop", "Invasion", "Lure", "Quest", "Weather", "Nest", "Gym"].filter((hookType) => this.config.general[`disable${hookType}`]).map((hookType) => hookType.toLowerCase())`

PR #680, opened 2022-02-28
Multiple geofence support

jfberry:
Simple implementation, supports an array for geofence parameter


hunch-13:
so far this is working well for my multiple cities/geofences.

unsure if impactful to poracleWeb

jfberry:
poracleweb and reactmap integration won't be impacted.  I might just complete this change (so hot reload works) and push, and come back to the other bits - unless of course they would also be helpful to you

hunch-13:
ok, perhaps not the place to ask but I'm unsure how poracleweb gets its area list(s) as I do not see the other geofences so i wasn't sure if this somewhat breaks the integration with PoracleWeb

PR #702, opened 2022-04-11
Add process title

lenisko:
## Description

Add process title. And yes.... we can't set it to full name `PoracleJS` due to the [limit](https://nodejs.org/api/process.html#processtitle).

## Motivation and Context

Helps to track and identify correct `node` process and it's usage.

## How Has This Been Tested?

On prod and dev env.

## Screenshots (if appropriate):
```sh
$ ps aux | egrep "Pora|PID"
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
len      3331214 13.5  0.2 2002304 270524 ?      Sl   02:19   0:15 PoraJS
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #703, opened 2022-04-12
Update package.json target

lenisko:
See https://github.com/KartulUdus/PoracleJS/pull/702#issuecomment-1097021321


lenisko:
@jfberry I was sure that `node .` was running `/poracle.js` as entry point, but the target is `src/app.js`

TIL, thanks!

PR #749, opened 2022-07-30
Add support for more Telegram parse modes

Kinrre:
## Description
Telegram notifications support three different parse modes, MarkdownV2, HTML and Markdown (legacy) https://core.telegram.org/bots/api#markdownv2-style.

PoracleJS hardcodes the usage of the legacy mode, but it presents a problem, for example, it can not be underlined text in the notifications, so I have added that the user can use the parse mode it would like.

*Note:* MarkdownV2 requires lot of escaped characters like the dot '.', so its a problem for not rounded IVs.

## How Has This Been Tested?

HTML and Markdown works perfectly, MarkdownV2 has its problems as it is stated before.

I also updated the default DTS to specify the parse mode, so the users can know that this support has been included.

Example of underlined and italic text using the HTML parse mode.
html_telegram

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation.


jfberry:
Perhaps could regularise the setting to make sure it isn't one that the api is going to choke on
eg

```javascript
let parse_mode = 'Markdown'
switch((data.message.parse_mode ?? 'markdown').toLowerCase()) {
  case 'markdownv2': { 
    parse_mode = 'MarkdownV2';
    break;
  }
  case 'html': {
     parse_mode = 'HTML'
     break
   }
  default : 
     break
}

```

Kinrre:
I have been testing your suggestion and the parse mode for the API is case insensitive, so it is not needed to make that change no?

jfberry:
> I have been testing your suggestion and the parse mode for the API is case insensitive, so it is not needed to make that change no?

I didn't realise that, but in any case this restricts to supported types rather than having an error so still useful?

Kinrre:
Yeah, you are right, I haven't thought about that. I just added the part of code that restrict to supported types.

jfberry:
If this is all tested then and you are happy I will merge :-)

Kinrre:
Yeah, all good from my side (HTML + Markdown)! 😄 

MarkdownV2 has its problems as stated in the first message, but with HTML you can do everything legacy Markdown doesn't support.

jfberry:
I think HTML matches poracle's handlebars capability better anyway, escaping happens automatically for & etc with {{ }}

PR #767, opened 2022-09-22
Lowering default minCP for PvP

Astu04:
Current Ultra won't cover cases like rank1 Stunfisk Galarian with CP 2445 at level 50, Togedemaru with CP 2442 or Skarmory with CP 2383. 

Current Great won't cover cases like rank1 Chansey with CP 1418.


jfberry:
PRs only into develop please

Issue #771, opened 2022-09-29
Gym reactMapUrl does not support RDM webhook


Fabio1988:
https://github.com/KartulUdus/PoracleJS/blob/9ca5cf7550a00933f3f6b0f805fff2a5cc6b1c45/src/controllers/gym.js#L97

It uses `gym_id` only (MAD), RDM sends `id` and url can't be resolved

@jfberry it changed in fort rework of [RDM](https://github.com/RealDeviceMap/RealDeviceMap/pull/394), we discussed that, maybe we missed a spot.... The fallback is after the map url creation

https://github.com/KartulUdus/PoracleJS/blob/9ca5cf7550a00933f3f6b0f805fff2a5cc6b1c45/src/controllers/gym.js#L130

PS: rdmUrl is build the same way, copied from reactMapUrl logic

PR #772, opened 2022-09-29
GymId fix for Map Linking

ReuschelCGN:
fix for issue:
Gym reactMapUrl does not support RDM webhook #771
https://github.com/KartulUdus/PoracleJS/issues/771

Tested by Fabio & ReuschelCGN and works with MAD and RDM now.


Fabio1988 (approved):

Fabio1988:
Thanks running that now since 1 week. it fixed the linking for both cases :)

PR #778, opened 2022-10-13
fix autocreate mixing permissions between roles

na-ji:
## Description

When using the autocreate command, and putting multiple roles with different permissions in the channel definition, I noticed that the permissions were mixed between the different roles. For instance:
```json
          "roles": [
            {
              "id": "AAA",
              "view": true,
              "viewHistory": true,
              "send": true,
              "react": true
            },
            {
              "id": "BBB",
              "view": false,
              "viewHistory": false,
              "send": false,
              "react": false
            },
          ]
```
With this definition, when creating the channel, the role `BBB` will have the same permissions as the role `AAA`.

## Motivation and Context

The fix is quite easy. I just made it so that the permissions array (called `allowed` here) is reset between each role creation. 

## How Has This Been Tested?

I tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/discord/commando/commands/autocreate.js
+++ b/src/lib/discord/commando/commands/autocreate.js
@@ -95,10 +95,10 @@ exports.run = async (client, msg, [args]) => {
 
 			// add role permissions
 			let roleId
-			const allowed = []
 			if (channelDefinition.roles) {
 				const roleOverwrites = []
 				for (const role of channelDefinition.roles) {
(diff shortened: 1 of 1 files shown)


jfberry:
Please resubmit on develop branch and I will merge
Thank you for your contribution!

na-ji:
> Please resubmit on develop branch and I will merge

This should be good now. :)

PR #791, opened 2022-11-04
fix autocreate channels deny

Fabio1988:
Poracle's autocreate command did not respect the `false` values for permissions.

This change should make it possible to deny view for everyone role

tested and works as intended :)

### DIFF
--- a/src/lib/discord/commando/commands/autocreate.js
+++ b/src/lib/discord/commando/commands/autocreate.js
@@ -75,8 +75,13 @@ exports.run = async (client, msg, [args]) => {
 
 		let categoryId
 		if (template.definition.category) {
-			const category = await guild.channels.create(format(template.definition.category, args), { type: 'GUILD_CATEGORY' })
-			categoryId = category.id
+			const exists = await guild.channels.cache.find(type => type.type === 'GUILD_CATEGORY' && type.name === template.definition.category)
+			if (exists) {
+				categoryId = exists.id
+			} else {
+				const category = await guild.channels.create(format(template.definition.category, args), { type: 'GUILD_CATEGORY' })
+				categoryId = category.id
+			}
 		}
 
 		for (const channelDefinition of template.definition.channels) {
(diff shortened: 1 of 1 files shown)


jfberry:
I don't use this but I wonder if it should differentiate between undefined and false... undefined don't do anything, true - set, false deny?

Fabio1988:
True we should differentiate between undefined and false

I tested and it worked smoothly... Small hint how to differentiate between undefined and false?

```else if(!role.view)```

Does this only match false?

**Comment by jfberry** [comment]
> True we should differentiate between undefined and false
> 
> I tested and it worked smoothly... Small hint how to differentiate between undefined and false?
> 
> `else if(!role.view)`
> 
> Does this only match false?

javascript has the concept of 'truthy' and 'falsey' which is entirely unhelpful here.
!role.view will match both false and undefined
you have to check specifically
if (role.view === undefined)

**Comment by Fabio1988** [comment]
I used now `role.view === false` should do the same and is nicer to read..

also added category exists already check :)

File: src/lib/discord/commando/commands/autocreate.js [hunk 1]
```diff
@@ -75,8 +75,13 @@ exports.run = async (client, msg, [args]) => {
 
 		let categoryId
 		if (template.definition.category) {
-			const category = await guild.channels.create(format(template.definition.category, args), { type: 'GUILD_CATEGORY' })
-			categoryId = category.id
+			const exists = await guild.channels.cache.find(channel => channel.type === 'GUILD_CATEGORY' && channel.name === template.definition.category)
```
Fabio1988, line 78: somehow this does not give any result, because it never finds existing categories, it always creates a new one with identical name

jfberry: The original code is intended to create new categories.  You will notice you are comparing to the raw string name not an expanded one as well

Fabio1988: Yeah it depends on the template used. I used a template with a variable category, I define it as param :)

That's why I thought about fetching existing categories first, but I will remove that part.... The deny is fixed and worked as intended :)

Fabio1988:
I removed the additional logic ,this PR only fixes the deny of discord permissions for now :)

PR #792, opened 2022-11-14
lift up pogo-data-generator Version to 1.8.4 in package.json

ReuschelCGN:
etc. fix for rocket lineup ...


TurtIeSocks:
Should squash before merging all of these commits :)

PR #794, opened 2022-12-06
Update Github Action Versions

Fabio1988:
Update because of deprecations


Fabio1988:
@jfberry this will also fix the docker images.. because the actual develop image is using axios 1.2.0....
The Image of this PR already contains 1.2.1 :+1:

PR #796, opened 2022-12-15
Provide tracking by monster size

jfberry:
`!track everything size:m-xxl` -- spread of size
`!track everything size:xxl` -- just xxl
`!track everything size:xxs` -- just xxs

Requires RDM size PR and size to be in webhook

New handlebars options: `{{sizeName}}` 

I'm not in a position to be able to test this so this is classed as experimental (but other people are telling me it works)


ChaosEmpire:
after Fix, all is working as expected

PR #828, opened 2023-03-21
Fix JSON parsing in koji.js

### LINKED ISSUE #827: Koji config is not readable from docker
https://github.com/KartulUdus/PoracleJS/blob/9224242e06befb4e0aa0c369683a1355ee336edf/src/util/koji.js#L17

If I remove that check manually bearer token can be read from config. No ugly json, can be parsed successfully without that check.

Fabio1988:
## Description
It can happen that JSON files have a special encoding character at the beginning, this check would prevent loading the json. we already use try/catch here

## Motivation and Context

I want to use Koji with Poracle (both in docker)
fixes #827 

## How Has This Been Tested?

Bearer token is now detected after that change

## Types of changes

## Checklist:

### DIFF
--- a/src/util/koji.js
+++ b/src/util/koji.js
@@ -14,9 +14,7 @@ const parseConfigSafe = () => {
 					.readFileSync(resolve(__dirname, '../../config/local.json'))
 					.toString(),
 			)
-			if (string.startsWith('{') && string.endsWith('}')) {
-				config = JSON.parse(string)
-			}
+			config = JSON.parse(string)
 		}
 		return config
 	} catch (_) {


jfberry:
Really this shouldn't be happening here at all but rather in the load geofence where proper config loader has already been called

TurtIeSocks:
I'm fine to bring this process internally, but I would need help with it as I couldn't get it to work nicely with the rest of the codebase.

TurtIeSocks (approved):

Fabio1988:
@jfberry could we merge that fix until we find a better solution pls? :)

jfberry:
The fix you have taken out was introduced because it was crashing

Fabio1988:
> The fix you have taken out was introduced because it was crashing

weird it's not crashing for me with that change. so we have several cases. and we definitely have to fix/rework the way how we use koji

Fabio1988:
reopen the ticket again

PR #829, opened 2023-03-22
Add YOURLS Support, Using Styles in Config, and Getting Street Intersections from Geonames

auisgold:
## Description
Added YOURLS as a link shortening option, use yourls in config with Provider URL being your URL for YOURLS and Provide Key being YOURLS signature

Made change so that you can put time of day styles in config rather than through tileserver, this way you can just put "#(style)" in your template rather than praying you get the if else string right in your template. 

Added the ability to get street intersections from Geonames, pretty sure none of the other geocoding providers currently offer this and this is probably one of the more useful ways a user can quickly determine if they want to get something or not. 

## Motivation and Context
Thinking of moving over from WDR as I don't really have much time anymore to keep maintaining it and wanted to add some of the features that I was using in WDR into Poracle. 

## How Has This Been Tested?
Currently Haven't Tested any of this on Poracle. Only "testing" that has been done is on the requests to YOURLS and Geonames  as the request code is essentially the same as it was in WDR. 

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x ] My code follows the code style of this project.
- [ x ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly
Wiki would need updated with new intersection feature. Configs have been updated accordingly to reflect the new features.


jfberry:
Thank you for your contribution!  I really appreciate it.

These are all worthwhile changes, but it would be easier to review if they were three separate PRs - especially as we'll need to find testers for each one if you haven't been able to test the changes.
I have only flicked through the files and can see a few items
* by dropping the flags for dawn, dusk, etc you would break existing users styles;
* the setting of the styles could probably be done in the same (shared) code as the dawn/dusk calculation - or other shared code
* I'm not actually sure what the intersection thing is, but it probably doesn't belong in controller.js as it seems more of a utlity function; .. config is also in snake_case not camelCase

auisgold:
Thanks for the feedback, I'll start working on making these into separate PRs and fixing up some things:

- The setting of styles needs access to configs so it can't be done in nightTime.js, haven't looked at the code extensively so I just put things where they would work, but if you know of a better spot then I can move it. 
- Intersection just returns the closest intersection of streets to a coordinate (i.e. 1st Street and Elm Street), which I beleive the current geocoders don't do. I put it after getAddress seeing it as an extension to geocoding, but again if you think there's a better place for it I can move it.

jfberry:
replaced by seperate PRs (thanks)

PR #832, opened 2023-03-22
Get Street Intersections from Geonames

auisgold:
## Description
Added the ability to get street intersections from Geonames, pretty sure none of the other geocoding providers currently offer this and this is probably one of the more useful ways a user can quickly determine if they want to get something or not.

## Motivation and Context
People know street intersections very well, so this helps them to know if it's worth it or not to go after a mon at quick glance. 

## How Has This Been Tested?
Use {{intersection}} in DTS in order to get intersection of streets. Make sure to use multiple users for geonames if you will hit rate limit of 20k per day

## Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ X ] My code follows the code style of this project.
- [ X ] My change requires a change to the documentation.
- [ X ] I have updated the documentation accordingly.
Config option added to use multiple users for Geonames to not reach rate limit.

### DIFF
--- a/src/controllers/controller.js
+++ b/src/controllers/controller.js
@@ -7,6 +7,7 @@ const TileserverPregen = require('../lib/tileserverPregen')
 const replaceAsync = require('../util/stringReplaceAsync')
 const HideUriShortener = require('../lib/hideuriUrlShortener')
 const ShlinkUriShortener = require('../lib/shlinkUrlShortener')
+const axios = require('axios')
 
 const EmojiLookup = require('../lib/emojiLookup')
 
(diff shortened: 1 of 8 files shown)


src/controllers/controller.js
```diff
@@ -378,6 +379,36 @@ class Controller extends EventEmitter {
 		return addr
 	}
 
+	async getIntersection(latitude, longitude) {
+		if (this.config.geocoding.intersection_users == [""]){
```
jfberry, line 383: needs updating for camel case
should also live through not being present.
perhaps
`
if (!this.config.geocoding.intersectionUsers || !this.config.geocoding.intersectionUsers.length) ?

src/controllers/controller.js
```diff
@@ -378,6 +379,36 @@ class Controller extends EventEmitter {
...
+			return 'No Intersection'
```
jfberry, line 384: would need to provide a translation lookup (function could have translator passed in)

src/controllers/gym.js
```diff
@@ -178,6 +178,8 @@ class Gym extends Controller {
 
 					require('./common/nightTime').setNightTime(data, conqueredTime)
 
+					data.intersection = await this.getIntersection(data.latitude, data.longitude);
```
jfberry, line 181: looks like getIntersection returns an array... how are you expecting the DTS to be used?

auisgold: Was using an array for WDR, just changed it to return a string so it should be good.

src/controllers/controller.js
auisgold: Was able to fix everything else, not really sure how to get the translation in.

jfberry:
This looks ready to go - please DM me in discord to confirm status with regard to testing

PR #838, opened 2023-04-13
Generate questString using quest_title from RDM

auisgold:
## Description
Creates questString based on quest_title from RDM rather than quest_conditions

## Motivation and Context
This change makes it so questString returns a more detailed description of the task that needs to be done rather than a generic one. 

## How Has This Been Tested?
I have tested this and it works without any errors. Users are happy as well that the task description is more detailed than before

## Types of changes
- [ X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ X ] My code follows the code style of this project.


auisgold:
This now has support for: de, es, fr, it, and ru locales.

ReuschelCGN:
PR works good with german translations, maybe you could cleanup changes in files:
src/controllers/monster.js
src/util/generateData.js
tileservercache_templates/poracle-area.json
tileservercache_templates/poracle-distance.json

Then it could be merged into develop!

jfberry:
sorry for the delay.  I will merge develop to master later this week, then can take these changes (after I review). I've been really slow lately, for which I apologise :-)

src/controllers/quest.js
```diff
@@ -386,175 +382,21 @@ class Quest extends Controller {
...
-			str = str.replace('{0}', item.target)
-			if (item.target === 1) {
-				str = str.replace('(s)', '').replace('1 ', 'a ').replace(' a times', '').replace('friends', 'friend')
-			} else {
-				str = str.replace('(s)', 's')
+		if (item.title) {
+			item.quest_title = item.title
+		}
+		const questinfo = `quest_title_${item.title}`
+		const questTitle = translations[this.config.general.locale].questTitles
```
jfberry, line 390: This is translating once? Surely this should be translating based on the human's current settings?
Should also be a default for the unknownQuestString if one is not configured
GameData should also be in this.GameData so suspect translations should pop through as this.GameData.translations and not need importing directly

auisgold:
Changed things to translate based on human's settings and switched to using this.GameData. Set unknownQuestString to 'Unknown Task'.

jfberry:
Would appreciate someone testing this with MAD before merge (but looks good to me)

fosJoddie:
Works fine with MAD backend.

Tested with ENGLISH language

PR #841, opened 2023-05-24
steelgrunt emoji fix/patch

ReuschelCGN:
tested since month in my server...

### DIFF
--- a/src/controllers/pokestop.js
+++ b/src/controllers/pokestop.js
@@ -231,19 +231,20 @@ class Invasion extends Controller {
 							data.gruntRewards = ''
 							if (data.gruntTypeId in this.GameData.grunts) {
 								const gruntType = this.GameData.grunts[data.gruntTypeId]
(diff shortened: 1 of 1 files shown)


src/controllers/pokestop.js
```diff
@@ -237,8 +237,8 @@ class Invasion extends Controller {
 								if (!data.genderDataEng) {
 									data.genderDataEng = { name: '', emoji: '' }
 								}
-								if (this.GameData.utilData.types[gruntType.type]) {
-									data.gruntTypeEmoji = translator.translate(this.emojiLookup.lookup(this.GameData.utilData.types[gruntType.type].emoji, platform))
+								if (this.GameData.utilData.types[gruntType.type === 'Metal' ? 'Steel' : gruntType.type]) {
```
jfberry, line 240: Looks like the type is also used for colour below,
maybe
`const t = gruntType.type === 'Metal' ? 'Steel' : gruntType.type` then use t in the following code?

PR #842, opened 2023-05-26
Fix raid/egg Everything Command

Kneckter:
The last code change in #839 added a `+` to the everything command, which made it stop at level 9. This fix allows the command to track the whole array of raid levels.

### DIFF
--- a/src/lib/poracleMessage/commands/egg.js
+++ b/src/lib/poracleMessage/commands/egg.js
@@ -58,7 +58,7 @@ exports.run = async (client, msg, args, options) => {
 			else if (element === 'valor' || element === 'red') team = 2
 			else if (element === 'mystic' || element === 'blue') team = 1
 			else if (element === 'harmony' || element === 'gray') team = 0
-			else if (element === 'everything') Object.keys(raidLevels).forEach((x) => levelSet.add(+x))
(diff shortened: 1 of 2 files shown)


jfberry:
Did you check this?
It works fine for me before.  The + means 'convert to integer' in javascript.

Kneckter:
I did test it, I wasn't just reading code for the fun of it lol

Before
Screenshot_20230526-134733_Discord


After
Screenshot_20230526-134757_Discord

jfberry:
Before: 
image

I'm not sure why you would have a problem...

Kneckter:
I'm not sure but it did check the first 9. Maybe there is a different issue buy this worked

TurtIeSocks:
Could this be an issue related to editing existing tracks? In the first screenshot you aren't removing them first and each one says "unchanged"

Kneckter:
That is what I was thinking as an alternative issue. I'm just not home yet to recheck.

Kneckter:
I'm not sure what caused it but I'm closing this if it's not needed.

jfberry:
I wouldn't want you to close it if there is genuinely a problem, but I am not sure it is the one you thought it was. I need to rework these functions anyway as I notice that they incorrectly imported values that are passed as context but it won't change the sematics

PR #844, opened 2023-05-28
Support golbat fort update webhooks

jfberry:
Golbat provides fort update webhooks that show when a fort (pokestop or gym) has changed in game - ie moved, been added or removed.
This provides per user tracking for forts.
It's a work-in progress and I'll update the description with DTS values when I settle on them, but early comments are welcome


src/controllers/fortupdate.js
```diff
@@ -0,0 +1,232 @@
...
+
+				const language = cares.language || this.config.general.locale
+				const translator = this.translatorFactory.Translator(language)
+				let [platform] = cares.type.split(':')
+				if (platform === 'webhook') platform = 'discord'
+
+				// full build
+				data.name = translator.translate(data.nameEng)
+				data.formName = translator.translate(data.formNameEng)
+				data.shinyPossibleEmoji = data.shinyPossible ? translator.translate(this.emojiLookup.lookup('shiny', platform)) : ''
```
ccev, line 190: these have to go, specifically line 188 overwrites the fort name

jfberry: Thanks, have quickly removed (they are obviously cut and paste nonsense!)

PR #845, opened 2023-06-01
questStringRaw translation for RDM systems with WatWowMap/pogo-translations

ReuschelCGN:
**needs testing but works with german translations**

could be switched on in config with "useRDMQuestString": true, and didnt changed the old way
fetch all locales from: 
`https://raw.githubusercontent.com/WatWowMap/pogo-translations/master/static/locales`

NO support for MAD sorry!


Fabio1988:
I would change config to `useQuestTitleString`. But as many don't know what's that about, I would call it `useIngameQuestString`.

Default to true :)

src/controllers/quest.js
```diff
@@ -248,10 +253,25 @@ class Quest extends Controller {
 
 				const language = cares.language || this.config.general.locale
 				const translator = this.translatorFactory.Translator(language)
+				const questI18l = require(path.join(__dirname, `../util/locale/i18n_${language}.json`))
```
Fabio1988, line 256: `questI18n` 👍

config/default.json
```diff
@@ -72,6 +72,8 @@
      "defaultTemplateName": 1,
     // library of custom DTS templates
       "dtsDictionary": {},
+    // Queststring i18l translation for RDM systems
```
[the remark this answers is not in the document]
ReuschelCGN: changed

src/controllers/quest.js
ReuschelCGN: changed

PR #850, opened 2023-07-06
add pokemon showcase

ReuschelCGN:
tracking e.g.
`!invasion showcase d1000 clean`

use in DTS something like:
`{{#if (eq displayTypeId 9)}}Showcase{{else}}Invasion{{/if}}`


src/controllers/pokestop.js
```diff
@@ -147,6 +145,13 @@ class Invasion extends Controller {
 				data.gruntName = data.displayTypeId && this.GameData.utilData.pokestopEvent[data.displayTypeId] ? this.GameData.utilData.pokestopEvent[data.displayTypeId] : ''
 				data.gruntType = data.displayTypeId && this.GameData.utilData.pokestopEvent[data.displayTypeId] ? this.GameData.utilData.pokestopEvent[data.displayTypeId].toLowerCase() : ''
 				data.gruntRewards = ''
+				if (data.displayTypeId === 7) {
```
jfberry, line 148: These should be in util.json - see plenty of other examples.
Maybe the color and an emoji reference should be there?

ReuschelCGN: like this?
im actually working on it 
		"3": {
			"name": "Instinct",
			"color": "FFDE00",
			"emoji": "team-instinct"
		},

jfberry: Yes, exactly

jfberry:
Let me know when ready for review

ReuschelCGN:
`!incident` parallel to `!invasion` -> works like expected
`util.json change` -> works like expected also in DTS
but it looks like, when a showcase (runs 3 days) message is tracked with **clean** command, the message is deleted in discord after 24h.... no idea why!

we should also wait until PR#838 is merged !?!

PR #851, opened 2023-07-19
feat: increment fails on webhook 404

lenisko:
## Description

Make use of `human.fails` and increment a value upon webhook 404.

## Motivation and Context

Easier to remove deleted channels but staled webhooks

## How Has This Been Tested?

Production.

## Types of changes

## Checklist:


jfberry:
If this column (which is unused as now) is to be used, surely it should also increment on bot failures (eg user permission problems) ?
Not sure how I feel about pushing db access into the queue runner but can’t immediately think of a simple alternative

lenisko:
Wasn't use case handled the other way? I believe they are getting locked after abuse. But sure, I'll take a look later.

Yeah... couldn't think about a better solution in this codebase. I was 98% sure you will mention it ,)

jfberry:
I think that exception is the right place, have you seen this work?

lenisko:
- [x] webhook
- [x] discord:channel
- [ ] discord:user

Working for those. I'll edit once I get occurrence on users.

jfberry:
I imagine it would work. It is easily testable by setting DMs to friends only then they will fail for a user

jfberry:
...this would be a great addition to !userlist as well to show fail numbers (although without it resetting it would only be an indicator of a past problem I guess)

jfberry:
These can be in a future PR. Would be easy to add a last_fail_date, or zero at midnight or something to make it more relevant but for now it's a good indicator at least.
I was just thinking out loud rather than these things being blockers before merge

lenisko:
:) It's ready to go then

lenisko:
:poke: @jfberry

PR #854, opened 2023-07-20
Fix tracked command for forts

Fabio1988:


jfberry:
They should call me half-a-job jabes
Thanks!

PR #862, opened 2023-08-20
Update apiMasterData.js

RagingRectangle:
Adds api response for all of MasterData for use with Chatot.


jfberry:
Thanks for the contribution. PRs should be against the develop branch. 
I’ll look at content later!

you could also fix the form bug in raids you highlighted

jfberry:
Do you still need this for chatot? If not, please close

PR #863, opened 2023-08-30
QuestsFix v4 - fallback lang to en when lang isnt in masterfile for Questtask

ReuschelCGN:
if user use language in poracle for dts, which isnt in masterfile, the questtask gen. fails - so:
check masterfile for available language and if it dosnt match to user language -> fallback to english

tested by dkmur in his multilang poracle

### DIFF
--- a/src/controllers/quest.js
+++ b/src/controllers/quest.js
@@ -247,7 +247,12 @@ class Quest extends Controller {
 				let [platform] = cares.type.split(':')
 				if (platform === 'webhook') platform = 'discord'
 				if (language !== this.config.general.locale) {
-					data.questString = await this.getQuest(data, language)
+					if (!Object.keys(this.GameData.translations).includes(language)) {
+						this.log.warn(`Missing language ${language}, fallback to english`)
+						data.questString = await this.getQuest(data, "en")
+					} else {
+						data.questString = await this.getQuest(data, language)
+					}
 				}
 
 				for (const monster of data.rewardData.monsters) {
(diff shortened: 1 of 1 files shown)


ReuschelCGN:
the way it is now?

for german translation it works but i have no chance to test it like above for users with language not in masterfile

src/controllers/quest.js
```diff
@@ -387,6 +387,10 @@ class Quest extends Controller {
 		if (item.quest_task && !this.config.general.ignoreMADQuestString) {
 			str = item.quest_task
 		} else {
+			if (!Object.keys(this.GameData.translations).includes(language)) {
```
jfberry, line 390: I think you could just do `!this.GameData.translations[language]` ?
since that would return undefined?

ReuschelCGN: this:
```
			if (!Object.keys(this.GameData.translations).includes(language)) {
				this.log.warn(`Missing language ${language}, fallback to english`)
				language = 'en'
			}
```

to:
`			language = !this.GameData.translations[language] ? 'en' : language`

and no log entry, that it fallback ???

jfberry: a log entry is probably not required so this would work well.
some people don't like ?: as it doesn't scan as well visually; I have no problem with it

jfberry:
beautiful, if you confirm working will merge

ReuschelCGN:
for german translation (no fallback) it works,  but actually i have no possibility to test with fallback

PR #866, opened 2023-09-10
update Nplumb dts example files

nileplumb:
Update my dts file examples


jfberry:
Please make pr based off develop

jfberry:
Yes looks ok will check later when not on mobile

PR #870, opened 2023-09-19
fix: fort update webhook

TurtIeSocks:
## Description

- Removes old references to c&p'd nest code
- Sets `time` to current time

## Motivation and Context

Valid time and fix c&p bugs

## How Has This Been Tested?

Currently untested 
## Types of changes

## Checklist:


ReuschelCGN:
https://github.com/TurtIeSocks/PoracleJS/blob/f07c85725db0645ca78b9aed8592e6dc756cc14e/src/controllers/fortupdate.js#L208

think this could also be remove

ReuschelCGN:
removing Line: 100 makes probs with Lines: 255-258

jfberry:
Removing tth will cause the senders to fail if clean specified at very least

Fabio1988:
Should we even allow to `clean` those hooks? Imo this does not make sense, right?

jfberry:
> Should we even allow to `clean` those hooks? Imo this does not make sense, right?

Even if you don't allow clean in the track command the tth values need to be there for the sending code not to crash

PR #872, opened 2023-10-09
Implement rtree for faster geofence matching

jfberry:


Fabio1988 (approved):

Fabio1988:
running since few days now without any issues. looks promising, thanks a lot :)

PR #873, opened 2023-10-10
Fastify5 deprecation fix ... tested and works

ReuschelCGN:
runs in prod. server since several weeks


Fabio1988:
Thanks a lot, this fixed 

```
(node:18) [FSTDEP016] FastifyDeprecation: You are accessing the deprecated "request.routeConfig" property. Use "request.routeOptions.config" instead. Property "req.routeConfig" will be removed in `fastify@5`.
```

PR #875, opened 2023-10-13
Reduce memory usage by disabling pokemon cache

jfberry:
Include `disablePokemonCache":true` in tuning section to disable pokemon cache.

The impact of this will be that poracle uses a lot less memory.  But it relies on the fact that your sender of pokemon is not spamming poracle with unnecessary duplicate pokemon data.  Golbat, funnily enough, is quite a good source of only relevant changes.


TurtIeSocks (commented):
~16 hours in, RAM usage cut by 30-40%, no reports of dupes from Golbat. 🚀

PR #886, opened 2024-01-02
Poor Man's IV Queue

kbtbc:
EDIT SINCE CLOSED:
  THIS PR WORKS* _ FURTHER DEV & MERGE ON HOLD _ SEE COMMENTS

**NEW DATABASE TABLE REQUIRED** _ SEE COMMENTS
---------------original--------------
When an MonsterNoIV alert is triggered by a user, a scout is sent to Dragonite to scan the location. 

## Description
Adds configs for Scout

## Motivation and Context
Lost IV Queue support

## How Has This Been Tested?
Works for Me

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

Being new to js, this is probably wrong for reasons I'm unaware, but it works -- thought I'd share if someone else might improve or find useful.


kbtbc:
It's been pointed out to me that Dragonite does not use an api secret.  So, if you want to try this and your servers are in a public environment, you would need to adjust to use Dragonite-Admin for security.

kbtbc:
welp, I don't think my dupe checker is working.   lastcoords must not persist in state?    not sure how to solve

kamieniarz:
@kbtbc do you mind adding pokemon id to each shout log? Would be easier to track successful results and their total time

jfberry:
I will review later but pr will need to be rebased on develop before we can consider merging it. Thanks!

kbtbc:
[kbtbc](https://github.com/kbtbc) changed the base branch from master to develop [yesterday](https://github.com/KartulUdus/PoracleJS/pull/886#event-11371534344)

kbtbc:
@jfberry - [kbtbc](https://github.com/kbtbc) changed the base branch from master to develop [yesterday](https://github.com/KartulUdus/PoracleJS/pull/886#event-11371534344)

kbtbc:
The issue is my duplicate variable doesn't seem to keep state.  So, I was thinking I might need to change the location of this function or figure out how to make a variable state persist here -- both I have yet to figure you :-/

kbtbc:
PR updated to add option to scan a cluster around lat/lon to help pickup pokemon data that might not be at cell location.  
To use set new config option: ` "scoutClusters" : true` 

Warning:  if you have a lot of noIV alerts, this might cause issues.   Most of this is untested except on my own setup.

src/controllers/monster.js
```diff
@@ -885,6 +888,114 @@ class Monster extends Controller {
 						}
 
 						const templateType = (data.iv === -1) ? 'monsterNoIv' : 'monster'
+						
+						sendscout: if (data.iv === -1 && this.config.general.scoutURL) {
```
jfberry, line 892: This is not a loop, so labelling and break below will not do anything.

src/controllers/monster.js
```diff
@@ -885,6 +888,114 @@ class Monster extends Controller {
...
+									'X-Dragonite-Secret': this.config.general.dragoniteSecret
+									}
+							const url = this.config.general.scoutURL
+							
+							// Try avoid sending dupes, check current against last used:
+							if (typeof lastcoords === 'undefined' )  {
+								   const lastcoords = coords	// lastcoords not yet defined, set value
+								} else if (lastcoords === coords) {
+								   this.log.info(`[SCOUT] Skipping duplicate attempt with ${coords}`)
+								   break sendscout;
```
jfberry, line 906: I think this will break from the wrong place; instead you should be setting a flag and checking it later.

However, there is a broader problem here in that this code runs on <n> threads, and so whatever lastcoords is is only for this thread and so this duplicate elimination is unlikely to succeed in this form

src/controllers/monster.js
```diff
@@ -885,6 +888,114 @@ class Monster extends Controller {
...
+								var currentAngle = 0;
+									for(var i=0; i < 6; i++)
+									{
+										// X2 point will be cosine of angle * radius (range)
+										var x2 = data.latitude + Math.cos(currentAngle) * ((70*1.732)/1000);
+										// Y2 point will be sin * range
+										var y2 = data.longitude + Math.sin(currentAngle) * ((70*1.732)/1000);
+										const clusterCoord = JSON.stringify([[parseFloat(x2.toFixed(5)),parseFloat(y2.toFixed(5))]])
+											try {
+												this.log.info(`[SCOUT] ${data.encounter_id}: ${monster.name} CLUSTER posting to ${url} with ${clusterCoord}`)
```
jfberry, line 959: If you are posting to scout, you can post a list of co-ordinates in one call, and they will all be added to the queue at once.  If you want to hit multiple destinations you should build up an array (with .push) and send all in one go

src/controllers/monster.js
kbtbc: In my googling, labeling and breaking out of a conditional is acceptable.  (and it does seem to work).  
my ref:   <https://stackoverflow.com/questions/4851657/call-break-in-nested-if-statements>

kbtbc: The 'broader problem' of determining how to make lastcords (or any varable) persist outside of the thread is where I am stuck and can use more hints.

kbtbc: Please keep in mind, my programming experiences is near zero here and so I'm learning as I go.

src/controllers/monster.js
kbtbc: Originally, I had this direction - but got stuck trying to figure out how to push the array

kbtbc:
Also, my calculations for cluster points are wrong...  still working on that :-)

kbtbc:
My cluster calculations are not perfect, and it might  be because I'm not accounting for the non-spherical nature of earth(?) ... for example, my cluster (the 6 surrounding target location here) -- you will notice that they are not centered correctly.   This is as close as I was able to get: 
image

src/controllers/monster.js
kbtbc: How to prevent duplicates in a multi-threaded?  In searching I came up with using a lock file somehow or adding a redis module and using redis to keep state variables -- seems not quite suitable?   I can write the coords to the database in a new table and check there?

kbtbc: My duplicate noIV hits are extremely low, and so the amount of effort to solve preventing dupes I think is beyond me...

kbtbc:
How to prevent duplicates in a multi-threaded?  In searching I came up with using a lock file somehow or adding a redis module and using redis to keep state variables -- seems not quite suitable?

kbtbc:
Added database table `scout` to use for duplicate checks.   To setup, with your poracle database:
``` CREATE TABLE `scout` (
  `id` int NOT NULL,
  `lastcoords` varchar(30) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  PRIMARY KEY (`id`); ```
Then add one record: 
``` INSERT INTO scout (id, lastcoords) values (1, 'none'); ```   

After creating this, Poracle will query the database for the most recent scout request and avoid sending if it's the same, preventing duplicates.

File: src/controllers/monster.js (hunk 1 above)
Comment (kbtbc) [reply src/controllers/monster.js:893]: This contradicts the docs.

**Comment by jfberry** [comment]
> Added database table `scout` to use for duplicate checks. To setup, within your poracle database: `` CREATE TABLE `scout` ( `id` int NOT NULL, `lastcoords` varchar(30) COLLATE utf8mb4_unicode_ci DEFAULT NULL, PRIMARY KEY (`id`); `` Then add one record: `INSERT INTO scout (id, lastcoords) values (1, 'none');`
> 
> After creating this, Poracle will query the database for the most recent scout request and avoid sending if it's the same, preventing duplicates.

A better approach would be to use the existing inter-process communication to send the scout request to the main (since) thread, and do the de-dupe there from a short lived in-memory cache.

**Comment by kbtbc** [comment]
> A better approach would be to use the existing inter-process communication to send the scout request to the main (since) thread, and do the de-dupe there from a short lived in-memory cache.

Beyond my scope of abilities/interest

**Comment by jfberry** [comment]
> > A better approach would be to use the existing inter-process communication to send the scout request to the main (since) thread, and do the de-dupe there from a short lived in-memory cache.
> 
> Beyond my scope of abilities/interest

Afraid I won't merge a db based solution for this. So either remove the dupe checking, or do it in memory as I suggest. I'm happy to give you some additional guidance on where to look for similar code etc

**Comment by kbtbc** [comment]
You can close this then, I think I'm done -- works good for me.

**Comment by kbtbc** [comment]
...and we are back!

No more dupe checking needed!

**Comment by kamieniarz** [comment]
Any chance this could be pushed to develop? ;) I'm running this PR for over a week and I haven't seen any issues. Lots of pros tho ;)

File: config/default.json [hunk 4]
```diff
@@ -243,8 +243,23 @@
...
+	  "dragoniteTimeout": 10000    // ms timeout for scout tuning
```
jfberry, line 258: This should be in the tuning section, and I generally hide these timeouts to only help people if they are experiencing problems (this should never need varying)

config/default.json
```diff
@@ -243,8 +243,23 @@
...
   },
   //
-  // Configure your discord bot(s)
+  // Configure Poracle to send scouts to Dragonite
+  // This is triggered only by users monsterNoIV alerts (not all NoIV)
+  // Leave blank to keep disabled
+  //
+  "scouts": {
+  	  "scoutURL": "",              // http://<ip>:<port>/scout
+	  "dragoniteSecret": "",       // Secret for drangonite admin api
```
jfberry, line 252: Is this verified working via the admin proxy? If so, it should be clear here when this needs to be used (ie only if routing via admin)

src/controllers/monster.js
```diff
@@ -602,6 +603,95 @@ class Monster extends Controller {
...
 
+					//sendscout by bigfun
+					data.scoutResult = ''
+					sendscout: if ((data.iv === -1 && this.config.scouts?.scoutURL && (data.seenType != 'wild' || (data.seenType =='wild' && this.config.scouts?.scoutWild)) && !this.config.scouts?.scoutBlackList.includes(data.pokemonId)) && (this.config.scouts?.limitScoutRarity <= data.rarityGroup || data.rarityGroup == -1 && this.config.scouts?.limitScoutRarity < 2)) {
+						this.log.info(`[SCOUT] Readying ${monster.name}(${data.pokemonId}) rarityGroup: ${data.rarityGroup} - limitScoutRarity: ${this.config.scouts?.limitScoutRarity} - Clusters: ${(this.config.scouts?.scoutClusters && data.seenType == 'cell')}.`)
+
+						const coords = []
+						coords.push([data.latitude,data.longitude])
+						if (this.config.scouts?.scoutClusters && data.seenType == 'cell') {  //Cluster Scans w/ better math thanks to @RPOT
+							data.scoutResult = 'Cluster '
```
jfberry, line 614: Overall this looks like lint has not been applied.  Please 'npm run lint' and see

src/controllers/monster.js
```diff
@@ -602,6 +603,95 @@ class Monster extends Controller {
...
+							data.scoutResult = data.scoutResult + result.data
+							this.log.info(`[SCOUT] ${data.encounter_id}: ${monster.name} Scout sent successfully!`)
+							
+						} catch (error) {
+							if (error.response) {
+								this.log.warn(`[SCOUT] Scout Error - Got ${error.response.status}. Error: ${error.response.data ? error.response.data.reason : '?'}.`)
+							} else {
+								this.log.warn(`[SCOUT] Scout NoResponse Error: ${error}.`)
+							}
+							return null
```
jfberry, line 690: We shouldn't be returning null in any of these error conditions - these will be abandoning the whole monster processing in case of a scout error?

src/controllers/monster.js
kbtbc: I am unable to `npm run lint` in my windows environment:
```
> npm run lint

> poracle@4.8.0 lint
> ./node_modules/.bin/eslint ./src --fix

'.' is not recognized as an internal or external command,
operable program or batch file.
```

config/default.json
kbtbc: This is untested, and I clarified in PR

src/controllers/monster.js
kbtbc: I've removed the returns and used a flag instead.

src/controllers/monster.js
kbtbc: update: linted

kbtbc:
Thank you @jfberry,  I think I've managed to address most of your recent notes.   

TIL: lint 😛   ... & I did get in working on Windows calling eslint directly ` npx eslint .\src\controllers\monster.js --fix` and was able to fix most lint errors except this one:  `Move function declaration to function body root  no-inner-declarations`  
https://github.com/KartulUdus/PoracleJS/pull/886/files#diff-31c72fd3572ce36ba8f4d27394dda49732e705adc42906717c56e82871825c18R614
These are the cluster calculation functions and I tried moving to root of monsters.js but data.lat/lon not yet declared error.  I tried elsewhere, just before the scout conditional, also with no luck.    Should I add the functions to app.js ?   ...  or how should I solve?  (It seems odd to have to declare function outside of the only conditional where they might be used.)

kbtbc:
I think the PR is ready to merge into develop at this point.  I have further ideas, but will probably take a while and this seems pretty solid & working foundation.   Also, I must admit @jfberry was right.  You cannot use labelled nested conditionals despite what the docs say.   My apologies for doubting you, sir.

PR #919, opened 2025-02-13
feat: API-less AccuWeather

Mygod:
## Description

Provide a fallback AccuWeather handler in case no API keys are available. Based on: https://github.com/Mygod/reactmap-android/blob/b81ea5251a771bd10f30e69cd5c57b6c7b64231f/app/src/main/java/be/mygod/reactmap/webkit/AccuWeatherDialogFragment.kt

Should we prefer to using the `web-api` endpoint instead of `geoposition/search` for obtaining `locationKey`? It achieves the same effect, is very efficient, and does not cost an API query.

This code is FOR EDUCATIONAL PURPOSES ONLY.

## Motivation and Context

Maintaining AccuWeather API keys are cumbersome. This fallback achieves the same effect without using any API keys, but with the following shortcomings:

* Wind speeds and wind gust speeds are rounded to integer in this interface so windy detection is slightly less reliable;
* Need to parse HTML for obtaining weather data so slightly less efficient;
* Probably less stable?

## How Has This Been Tested?

From logs: `Accuweather fallback forecast [GMT] 15:00 = 4  16:00 = 4  17:00 = 3  18:00 = 3  19:00 = 3  20:00 = 3  21:00 = 3  22:00 = 3  23:00 = 3`

No API keys configured.

## Types of changes

## Checklist:

### DIFF
--- a/src/controllers/weather.js
+++ b/src/controllers/weather.js
@@ -21,7 +21,7 @@ const accuWeatherWebApiOptions = {
 }
 const weatherForecastRegexp = /^\/en\/([^/]*\/[^/]*\/[^/]*)\/weather-forecast\/([^?]*)/
 const hourlyWeatherForecastRegexp = /^https:\/\/www\.accuweather\.com\/en\/([^/]*\/[^/]*\/[^/]*)\/hourly-weather-forecast\//
-const hourlyHtmlRegexp = /<div id="(\d+)" data-qa="\1" class="accordion-item hour".*?data-src="\/images\/weathericons\/(\d+).svg".*?(\d+) km\/h.*?(\d+) km\/h/s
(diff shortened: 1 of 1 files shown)


jfberry:
I am conflicted about this PR.  In a way it's convenient because users won't have to use the API any more (and setting up multiple api accounts is not best practice) - however scraping the web page is also not good practice and is fragile to change or other simple browser based protections

Mygod:
That's why this is just a fallback mechanism. You will never run this code if you keep your Poracle fed with enough API tokens. :)

KartulUdus:
This is a terrible idea!

An easy way to get banned off accuweather.
As an open source app, you are of course welcome to run your setup as you please, but I don't want users who run default settings on poracle to be blocked off any service because of it.
[closing]

PR #929, opened 2025-08-29
feat: initial maxbattle tracking in poracle

ReuschelCGN:
init PR for maxbattle tracking in poracle
(needs Golbat: https://github.com/UnownHash/Golbat/tree/stations-wh)
maxbattle tileserver templates, basic dts.json and poracle testdata are included
`!poracle-test max-battle gmax`
`!poracle-test max-battle level3`
`!poracle-test max-battle level1`

**IMPORTANT:** 
this PR needs testing and feedback specially for the icons (i use the uicons.js from turtlesocks in poracle) and telegram

#####   Maxbattle Tracking  #####
!maxbattle <MaxbattleBossName> d800 clean  ->  reports all "MaxbattleBossName" Maxbattles within 800 meters (d=distance in meters) of your set location and deletes the report after expiration
!maxbattle <MaxbattleBossName> d500 gmax clean  ->  reports all "MaxbattleBossName" gmax battles within 500 meters (d=distance in meters) of your set location and deletes the report after expiration
!maxbattle level<MaxbattleLevel>  ->  reports all Maxbattles with level "MaxbattleLevel" (possible Level: 1-8 where Level 6-8 are gmax Maxbattles)
!maxbattle remove <MaxbattleBossName>  ->  removes "MaxbattleBossName" from the Maxbattle tracking
#available variables are: <MaxbattleBossName>, level, move, gmax (only Gmax Battles), distance, clean

maxbattle


src/app.js
```diff
@@ -797,7 +796,27 @@ async function processOne(hook) {
...
 			}
+			case 'max_battle': {
+				if (config.general.disableMaxBattle) {
+					fastify.controllerLog.debug(`${hook.message.id}: MaxBattle was received but set to be ignored in config`)
+
+					break
+				}
+				if (!hook.message.poracleTest) {
+					fastify.webhooks.info(`max_battle ${JSON.stringify(hook.message)}`)
+					const cacheKey = `${hook.message.id}${hook.message.battle_end}${hook.message.battle_pokemon_id}`
```
jfberry, line 807: The cache will expire well ahead of the max battle most probably, so this may not provide any value whatsoever.
What causes golbat to resend a station webhook and do we care that this will generate a new alert?

ReuschelCGN: hook.message.id -> Station ID
hook.message.battle_end -> is eqal to end_time in golbat DB
hook.message.battle_pokemon_id -> if this change we get a new alert on same station

Why the cache should expire in normal use ahead of the max battle end?
But on Maxmondays or Events, it should gen. a new alert.

Fabio1988: Jabes meant that you put an element into cache, but the cache itself lasts for x minutes?! Maybe an hour? So cache lookup will almost never happen

Golbat is not sending any duplicate hooks, it only sends if there something changes, like battle end, or Pokémon etc.

jfberry: Yes, the cache lasts an hour (I think). And the max battle has a duration much longer than that, so updates within the period will not be eliminated by the cache

ReuschelCGN: ok so the cache could look like that:

`const cacheKey = `${hook.message.id}${hook.message.battle_pokemon_id}``

jfberry: Rephrasing my question - are there any duplicate webhook messages that you are trying to avoid passing into the alerting engine?
The cache here is simply used to avoid users getting duplicate alerts - if all conditions need to be passed through the cache is just a waste of memory.
This is why disabling the cache for pokemon is ok now - as golbat doesn't send webhooks for the same pokemon unless there is a significant change - whereas MAD (moreso) and RDM had a tendancy to duplicate information causing end user alerting pain

ReuschelCGN: > Rephrasing my question - are there any duplicate webhook messages that you are trying to avoid passing into the alerting engine?

Maxbattles have rotating moves and there are stationed mons which change, so i wanted to be sure, not to gen unnecessary alterts. But maybe im wrong with my thought.

Fabio1988: Max Monday would be such a case... For one hour there is a certain mon deployed... After that hour it changes back to the old one.... Golbat might send another hook
But the old mon can be already available for days, a one hour cache won't belp on that duplicate

Fabio1988: Can we increase the cache for max battles to 2h? So we would safely not send a duplicate webhook on the 7PM change? :)

ReuschelCGN: https://github.com/KartulUdus/PoracleJS/blob/1ab5309625b08bd31b8b7efd0114e909ed0955d0/src/app.js#L49

looks like its gen. 90min cache

Fabio1988: Hmm... As the first webhook could be sent the day before, we might still send another hook at 7PM... If I think about it, that's not that bad I guess

Do you have some notification examples for max monday or e.g. this event with Toxtricity?! :)

Fabio1988:
https://github.com/UnownHash/Golbat/pull/312 merged

jfberry:
Added to Poracle NG

PR #930, opened 2025-09-08
Multiple changes to API

lenisko:
feat(api): raid - support passing multiple levels and pokemon+forms
feat(api): query.silent to mute messages in tracking POST
feat(api): extend get/post with optional param profile_no overriding currently selected profile pull

## Description

Extended POST API used to create with non-breaking changes:
- egg: accept `level` as `int || []int`
- raid: accept `level` as `int || []int` and additional field `pokemon_form` which takes `pokemon_id, form` as objects.
- all tracking POST's when query `silent` arg is set, skip sending message to user, use `profile_no` query first then `current_profile_no` during creation
- all trackings GET's try to use `profile_no` query first then `current_profile_no`

## Motivation and Context

Same as accepting multiple gyms, some fields make sense to be specified in single call - `level`, `pokemon && form`. This makes sure we are pushing all changes in single HTTP request instead sending multiple and possible losing something on the way.

## How Has This Been Tested?

Tested on production using legacy and new API calls.

-

## Types of changes

## Checklist:


jfberry:
Don't see too much wrong with this, however you can indeed already send multiple levels by sending multiple rows (which is more in-line with the way the tracking is actually managed)

lenisko:
> Don't see too much wrong with this, however you can indeed already send multiple levels by sending multiple rows (which is more in-line with the way the tracking is actually managed)

Hmm... Since you don't see too much wrong, can it stay as it is?

Added a few more changes, some QOL and ability to GET/POST to profiles without swapping to them.

lenisko:
At this point, I don't care if we ever merge that. Development looks dead to me. 

Going to throw other things to bag on the way, I find them.

PR #936, opened 2026-01-03
Add dynamic RSVP raid message editing feature

warfy5:
Modified discordWorker and discordWebhookWorker to edit existing raid messages instead of sending new ones


jfberry:
Thank you - and sorry for the delay. This is a great feature idea, I've taken it into poracle-ng.
There you can do `!raid everything rsvp_only edit clean` to get a current list of rsvp-raids in your area that edit and clean themselves, for example

PR #937, opened 2026-01-23
Campfire

warfy5:
add support for campfire urls.

tested and working ok.


jfberry:
Are you able to split out the rsvp editing from this?
You've got at least three features in this PR

warfy5:
yes ok, seperating them out now. this should be just the campfire url now.

jfberry:
Added to Poracle NG

